Issue #1, opened 2025-07-15
feature: Allow file context to be passed as part of 'Explain this code'


gitfudge0:
### Did you check the docs?

- [x] I have read all the docs

### Is your feature request related to a problem? Please describe.

The feature "Explain this code" works great. However, passing only a piece of code loses context of what it is and where it's used.

### Describe the solution you'd like

Given how well opencode's file grepping works, passing file name with line numbers could work a little better because it passes the file location in context as well.

Instead of passing
```
func createEntry(with: Task)
// some code goes here for 10 lines
end
```

it's nicer to pass

```
app/service/task_service.go:L10-20
```

### Describe alternatives you've considered

I've had to pass the file name in the context after opencode processes the initial request and give more context

PR #2, opened 2025-07-15
feat: Send line numbers instead of code itself for improved context

gitfudge0:
Previously, the "Explain this code" feature would send only the selected code snippet, which often lost important context about where the code was located and how it was used.

This change updates the feature to send the file name along with the relevant line numbers (e.g., app/service/task_service.go:L10-20) instead of the code itself. By leveraging opencode's file grepping capabilities, this approach provides more precise context to the assistant, improving the quality of explanations and reducing the need for users to manually specify file locations.


nickjvandyke:
Nice, thank you @gitfudge0 !

nickjvandyke:
Closes #1

PR #7, opened 2025-07-24
use exact match for find instead of pattern matching

christo-auer:
`find` uses pattern matching instead of exact matching. This leads to problems when a path contains special characters used for expressions (like dashes). by using the given options, exact matching is used.


christo-auer:
will update PR according to your newest changes.

PR #8, opened 2025-07-24
exact match instead of pattern match

christo-auer:
`find` uses pattern matching. changed it such that uses an exact match. now it works with CWD which contain special characters (like dashes). also fixed missing return in `find_port`.


nickjvandyke:
Nice find, thank you for troubleshooting 😀

PR #15, opened 2025-08-06
fix(sse):The JSON decode error in event requests in the latest version of opencode

yunxi177:


nickjvandyke:
Thanks for the PR!

Can I ask what issues you had? I'm also on the latest `opencode` (on `brew` at least - `v0.3.131`) and haven't seen any. But maybe it takes a specific tool call or kind of response.

Regardless I knew my implementation could be cleaned up haha, will definitely take a look.

nickjvandyke:
Ah, I guess my implementation just happened to work (for me). I see now that the SSE spec separates events with blank lines, a more reliable way for us to buffer them

nickjvandyke:
Thank you! I simplified the conditionals in your code a bit - let me know if I made any mistakes 🙂

yunxi177:
> Thanks for the PR!感谢您的公关！
> 
> Can I ask what issues you had? I'm also on the latest `opencode` (on `brew` at least - `v0.3.131`) and haven't seen any. But maybe it takes a specific tool call or kind of response.我可以问问你有什么问题吗？我也使用最新的 `opencode`（至少在 `brew` - `v0.3.131`）上，但还没有看到任何。但也许它需要特定的工具调用或某种响应。
> 
> Regardless I knew my implementation could be cleaned up haha, will definitely take a look.不管怎样，我知道我的实现可以清理哈哈，一定会看看。

Maybe it’s related to my custom provider for opencode. I’ve found that in some cases, an event splits a complete JSON object into two separate transmissions, which causes JSON parsing to fail in the code.

nickjvandyke:
Ah that's a good point, I only use it with Copilot GPT-4.1. Well thanks for the find and fix!

yunxi177:
> Thank you! I simplified the conditionals in your code a bit - let me know if I made any mistakes 🙂谢谢！我稍微简化了您代码中的条件 - 如果我犯了任何错误🙂，请告诉我

Your simplified code works perfectly. Thank you for helping make the code clearer.

PR #21, opened 2025-08-13
Feat: a slightly better opencode sniffer

cds-amal:
## Summary

This PR enhances the robustness and debuggability of opencode port detection. 

## Changes

1. OS-aware port detection (64cfb51)

  - Multi-method fallback chain: Replaced single lsof-based detection with OS-specific methods that
  gracefully degrade
    - Linux: /proc filesystem → ss → netstat → lsof → HTTP probe
    - macOS: lsof → ps+lsof combo → HTTP probe
    - Windows/other: HTTP probe fallback
  - Performance improvements: Added 60-second cache to avoid repeated system calls
  - Broader compatibility: No longer requires lsof on Linux systems, works with different command
  variations

2. Enhanced debugging UX (1212f54)

  - Privacy-preserving diagnostics: Path sanitization utility protects user privacy when sharing
  debug logs
  - Multiple configuration methods:
    - OPENCODE_DEBUG environment variable for verbose logging
    - OPENCODE_PORT environment variable for manual port override
  - Interactive troubleshooting: Detection wizard and comprehensive diagnostics command
  - Documentation: Added troubleshooting guide to README with common issues and solutions

## Problem Solved

  Previously, pid and port detection failed for me on Ubuntu. It relied solely on lsof, which behaves differently across Linux 
  distributions and may not be installed. This caused the plugin to fail to find running opencode
  instances; much sad.

## Testing

Not enough :/

  - Tested on Ubuntu and Pop!_OS
  - Verified fallback chain works when individual commands are unavailable
  - Confirmed privacy sanitization doesn't leak sensitive paths

##  Breaking Changes
None - all changes are backward compatible with existing configurations.

## Acknowledgments

These improvements were developed with assistance from Claude (Anthropic) through iterative design and implementation discussions, focusing on cross-platform compatibility and user experience enhancements.
  
Co-authored-by: Claude <marvin_the_paranoid@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,65 @@ vim.api.nvim_create_autocmd("User", {
 | - | - |
 | `OpencodePlaceholder` | Placeholders in `ask` input |
 
+## 🔧 Troubleshooting
+
+### Port Detection Issues
+
+If `opencode.nvim` can't find your opencode server, use the diagnostic tool:
+
+```lua
+:lua require('opencode').show_diagnostics()
+```
+
+#### Manual Configuration
+
+1. **Environment Variable** (recommended for temporary fixes):
+   ```bash
+   export OPENCODE_PORT=37669
+   nvim
+   ```
+
+2. **Plugin Configuration** (permanent):
+   ```lua
+   require('opencode').setup({
+     port = 37669,  -- Your opencode port
+   })
+   ```
+
+3. **Finding the Port Manually**:
+   ```bash
+   # List opencode processes
+   ps aux | grep opencode
+   
+   # Find listening ports (Linux)
+   ss -tlnp | grep opencode
+   lsof -i -P | grep LISTEN | grep opencode
+   
+   # Find listening ports (macOS)
+   lsof -i -P | grep LISTEN | grep opencode
+   ```
+
+#### Debug Mode
+
+Enable verbose logging to see detailed detection attempts:
+
+```bash
+export OPENCODE_DEBUG=1
+nvim
+```
+
+This will show:
+- Which detection methods are being tried
+- Process discovery details
+- Port verification attempts
+- Exact failure points
+
+### Common Issues
+
+- **Multiple opencode instances**: The plugin looks for opencode running in Neovim's CWD. If you have multiple instances, it may pick the wrong one.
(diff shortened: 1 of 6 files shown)


cds-amal:
Hey @NickvanDyke, I was just testing this on a couple of my linux machines and see (conflict) you're working on the issue as well. Feel free to close this if yo want.

nickjvandyke:
Thanks for this, it's definitely a weak point in the plugin!

> you're working on the issue as well

So far I've only focused on `ps` usage - if `lsof` has issues too then this would definitely help 😀 I will take a look soon!

nickjvandyke:
@cds-amal did the PID-finding work for you in commit `c5659b5a0548ce7c254ad639a129fdb98d74c0dc` or earlier? After that is when I changed the PID-finding for `fish`'s sake. And I presume someone would've opened an issue if it was broken in Ubuntu all that time.

If it does work, I think I can just tweak the grep pattern a bit to still work with `fish`.

cds-amal:
> @cds-amal did the PID-finding work for you in commit `c5659b5a0548ce7c254ad639a129fdb98d74c0dc` or earlier? After that is when I changed the PID-finding for `fish`'s sake. And I presume someone would've opened an issue if it was broken in Ubuntu all that time.
> 
> If it does work, I think I can just tweak the grep pattern a bit to still work with `fish`.

I installed the plugin a few hours ago (0a4de5b), and it couldn't detect the opencode process/pid/port. I thought the `proc fs` approach would be a good bet.

nickjvandyke:
That commit a few hours ago was after the possibly-break `fish` accommodation 😀 https://github.com/NickvanDyke/opencode.nvim/commit/b9eac81d8ec28e9374a00b9ad816351a9f51b489

nickjvandyke:
I appreciate your effort but fortunately/unfortunately I fixed the issue with better use of `lsof` - https://github.com/NickvanDyke/opencode.nvim/commit/08bf33d8a0c381f6caa8c57f844899c41913b6e8. So far it seems reliable across distros, and `ps` was the real problem (now not used at all). But let me know if this implementation doesn't work for you 🙂 

I'm okay with the `lsof` dependency - it simplifies a lot of plugin code, most systems have it by default/it's an easy install, and users can still fall back to `opts.port`.

PR #29, opened 2025-08-30
feat: escaping Tmux navigator confilicts to cycle between panes

marcosktsz:
the following pr aims at providing tmux navigator support: right sending a command like "TmuxNavigateLeft" gets prompted into the chat instead of cycling through panes like expected

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -151,7 +151,20 @@ local defaults = {
         -- Make it easier to target for customization, and prevent possibly unintended "snacks_terminal" targeting.
         -- e.g. the recommended edgy.nvim integration puts all "snacks_terminal" windows at the bottom.
         filetype = "opencode_terminal",
-      }
+      },
+      on_buf = function(win)
+        local function setup_mux_nav()
+          if vim.env.TMUX and vim.fn.exists(":TmuxNavigateLeft") == 2 then
+            vim.keymap.set("t", "<C-h>", "<cmd>TmuxNavigateLeft<cr>", { buffer = win.buf, silent = true })
+            vim.keymap.set("t", "<C-j>", "<cmd>TmuxNavigateDown<cr>", { buffer = win.buf, silent = true })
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Thanks for raising this! Does this need to live in the plugin though? It seems to me like any tmux navigator users would want/already have these Terminal mode keymaps in their global config. I could be missing something but I don't see why they have to be set up specifically in the terminal's `on_buf`.

nickjvandyke:
Let me know if the above turns out to be incorrect!

varsis:
@NickvanDyke I'm experiencing this issue as well. 

What happens is the command is sent to the opencode instance instead of navigating as expected.  (See the screenshot) 
Screenshot 2025-09-09 at 21 12 29

I'm using this plugin: https://github.com/christoomey/vim-tmux-navigator locally.

nickjvandyke:
Thanks for the info @varsis. Maybe this is a quirk of `<cmd>` keymaps while in terminal mode? I wonder if it happens in a normal terminal buffer too - I don't see why it would only affect opencode's. If that plugin offers a Lua API, try that in your keymaps instead of commands. Ensure your keymap is set for terminal mode in the first place too - I see the plugin's readme only sets it for normal mode.

If the solution proposed by the PR is the only option, users can set it in their own `opts.terminal.win.on_buf`. It's too specific and inflexible for the default config.

varsis:
It seems the plugin I'm using is not working with terminal mode correctly. 

I added the following to my config and all works well now:
```
vim.keymap.set('t', '<C-h>', '<C-\\><C-n>:TmuxNavigateLeft<cr>')
vim.keymap.set('t', '<C-j>', '<C-\\><C-n>:TmuxNavigateDown<cr>')
vim.keymap.set('t', '<C-k>', '<C-\\><C-n>:TmuxNavigateUp<cr>')
vim.keymap.set('t', '<C-l>', '<C-\\><C-n>:TmuxNavigateRight<cr>')
```

marcosktsz:
thank you so much!

Issue #30, opened 2025-09-01
bug: Fix nil parameter in create_session function causing Malformed JSON error


yDog-1:
### Did you check docs and existing issues?

- [x] I have read all the plugin docs
- [x] I have searched the existing issues
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.11.3

### Operating system/version

Pop!_OS 22.04 LTS

### Opencode version (opencode -v)

0.5.6

### Describe the bug

Traditional code with nil parameter causes Malformed JSON error:
```json
{
  "data": {
    "message": "Error: Malformed JSON in request body"
  },
  "name": "UnknownError"
}
```

### Steps To Reproduce

1. require("opencode.client").create_session(port, callback)

### Expected Behavior

Session should be created successfully, returning a valid session object:

```json
{
  "id": "ses_6faca8612ffeIQ267Q7mCarWkL",
  "time": {
    "created": 1756729014765,
    "updated": 1756729014765
  },
  "title": "New session - 2025-09-01T12:16:54.765Z",
  "version": "0.5.6"
}
```

nickjvandyke:
Ah maybe this broke and I didn't realize because `create_session` is currently dead code on `main`

PR #31, opened 2025-09-01
fix(client): Fix nil parameter in create_session function error

yDog-1:
#30


nickjvandyke:
Thanks for finding and fixing! Also made me realize that `vim.fn.json_encode({}) == '[]'`, so I updated the other methods with `vim.empty_dict()` just in case!

PR #37, opened 2025-09-12
fix(client):Fix conditional logic to avoid unintended fallback when handle_sse returns nil or false

yunxi177:
In Lua, only false and nil are considered falsy, while all other values are truthy.
The current code uses the and … or … idiom:
```:lua
local response = is_sse and handle_sse(data) or handle_json(data)
```
This can cause unintended behavior: if is_sse is true but handle_sse(data) returns nil or false, the expression falls back to handle_json(data), which is not expected.

This PR replaces the expression with an explicit conditional to ensure the correct function is called and its result preserved:
```lua
local response
if is_sse then
  response = handle_sse(data)
else
  response = handle_json(data)
end
```
This makes the logic more robust and avoids unexpected fallbacks.

### DIFF
--- a/lua/opencode/client.lua
+++ b/lua/opencode/client.lua
@@ -81,7 +81,6 @@ local function curl(url, method, body, callback, is_sse)
       else
         response = handle_json(data)
       end
-
       if response and callback then
         callback(response)
       end


nickjvandyke:
oo that's a good catch, thank you!

PR #39, opened 2025-09-18
feat: add model option

carlosflorencio:
We can configure a global model for the `opencode` cli to launch with.


nickjvandyke:
Nice thank you, I didn't know they added this!

I'll merge this but FYI might generalize it to `opts.terminal.cmd` soon, so the plugin doesn't have to accommodate every possible flag

PR #52, opened 2025-10-09
Add example keymaps to README

Cih2001:
This is in fact the recommended way to setup key bindings in lazy nvim, because then it works with lazy loading (lazy nvim is aware of the key bindings set this way and can lazy load the plugin on demand)


nickjvandyke:
I appreciate the thought but this was intentional 🙂 I took care to lazy-load internally rather than offload it to every user - the plugin starts up in 0.4ms for me. And these keymaps etc. are plugin manager-agnostic, particularly with `vim.pack` on the horizon.

PR #62, opened 2025-10-21
feat: auto-find port on Windows

erroris3:
- Use TCP sockconnect instead of HTTP.
When a port is specified in the config, the current check hits the root path, which makes opencode return 404 and falsely fail. A TCP socket check should succeed if the opencode port is open, regardless of HTTP status.

- Auto-append --port flag.
Automatically inject the flag into the terminal command when a port is specified.

- Support find_server on Windows.
lsof isn’t available on Windows, so use a fallback with netstat + tasklist. For is_descendant_of_neovim, use wmic to walk the parent PID (including the grandparent) because opencode launched from Snack Terminal typically has a shell parent (e.g. PowerShell) Chaining to the grandparent to get the Neovim’s PID (shell -> opencode).

I've tested this on Windows and left the Unix part unchanged, so it should work fine

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -92,7 +92,7 @@ local defaults = {
   input = {
     prompt = "Ask opencode: ",
     -- `snacks.input`-only options
-    icon = "󱚣",
+    icon = "󱚣 ",
     win = {
       title_pos = "left",
       relative = "cursor",
@@ -102,11 +102,10 @@ local defaults = {
   },
   select = {
     prompt = "Prompt opencode: ",
-    -- `snacks.picker`-only options
-    picker = {
+    snacks = {
       preview = "preview",
       layout = {
-        -- Override default hiding of preview in `vim.ui.select`
+        -- preview is hidden by default in `vim.ui.select`
         hidden = {},
       },
     },


nickjvandyke:
Thanks for this! Would you mind PRing the first two changes separately? I'd love to get those simple workarounds in, and then review the windows-specific code when I have more time 🙂

erroris3:
Absolutely!  

https://github.com/NickvanDyke/opencode.nvim/pull/63#issue-3538440492

nickjvandyke:
Whenever you get the chance, could you merge `main`/drop the non-Windows changes that we merged in #63 ? Ty!

erroris3:
I have an idea. Instead of opening an opencode instance without the --port flag and finding the port afterward which causes a lot of lag when polling. I noticed a huge difference when specifying a port vs auto-finding one. Can I change it to randomly select a port that isn’t in use, so we don’t have to check whether it’s an opencode process or if it’s running through Neovim?

nickjvandyke:
It's a neat idea but then we couldn't find instances that are already running - a powerful feature.

How badly does port finding lag for you? I've never noticed much but maybe it's machine-dependent, or different with the windows implementation. `test_port` used to lag really badly for me but your PR fixed that.

nickjvandyke:
Not every part of the port-finding process is async right now, but could be. If you can identify the portion that lags the UI for you, we could migrate it to async.

erroris3:
Maybe its on me for the Windows auto-find. It requires many exec calls to find the opencode port, and when poll_for_port retries its lag out. I will try to find a better way and create new PR

nickjvandyke:
`exec` uses `io.popen` which runs the command in a new process, so it shouldn't lag the UI 🤔 At least it doesn't for me on MacOS. Let me know what you find!

I have thought to migrate `exec` to the newer `vim.system`, which is meant for async commands like this. Maybe explore that?

PR #63, opened 2025-10-21
fix(server): `opts.port` testing and auto-add --port flag to `terminal.cmd`

erroris3:


nickjvandyke:
Ty, this is great!

What do you think about making `terminal.cmd` *only* `string[]` with this change? More standard and it'd simplify your logic with no downsides (except a small breaking change). I'm afk though and can't remember if `snacks.terminal` accepts that, or we'd have to join it first.

erroris3:
you mean defaulting terminal.cmd to { "opencode" }?

nickjvandyke:
Yup!

erroris3:
I think we should keep snack.terminal.cmd as is, supporting both string and string[]. But I’m happy to change it if you want.

nickjvandyke:
No problem! On second look, do you think we could keep terminal.cmd as just `string`? I'm open to arguments for `string[]` but afaict it just complicates the interface and logic a bit without benefit (?). i.e. copilot is correct that we'd get duplicate `--port` arguments if `terminal.cmd` is a string that already contains it.

erroris3:
You are right. If we force it to be an array only, it would be much simpler

erroris3:
I will change it to array only tmr. Thx alot for the advice

nickjvandyke:
What do you think about string-only, so we don't have a breaking change? As a user I tend to find passing commands as strings to be simpler than arrays, but I could be missing an upside to the array format. I guess it makes parsing easier? But `cmd` is not a complex string anyway.

erroris3:
Thats better

nickjvandyke:
Sweet! Thanks for discussing 🙂

erroris3:
Sorry for late commit. here is the string only version

nickjvandyke:
I pushed a couple changes with docs to save you some back-and-forth 🙂

Thanks again for this! While testing the plugin prior to your changes I realized finding the port lags the UI and occasionally throws errors, so your fix is even more important than I thought 😁

PR #65, opened 2025-10-25
feat(server): auto-find port on Windows with async

erroris3:
- Remove polling for port when its not found the opencode to randomly assign a free port and return
- Made its all async now its no more UI lag.

**BUT** i giveup on finding the process cwd on windows so skip the check for that. Checking the parent should be enough (?).

### DIFF
--- a/lua/opencode/cli/server.lua
+++ b/lua/opencode/cli/server.lua
@@ -1,27 +1,70 @@
+local IS_WINDOWS = vim.fn.has("win32") == 1
+
 local M = {}
 
----@param command string
----@return string
-local function exec(command)
-  -- TODO: Use vim.fn.jobstart for async, and so I can capture stderr (to throw error instead of it writing to the buffer).
-  -- (or even the newer `vim.system`? Could update client.lua too? Or maybe not because SSE is long-running.)
-  local executable = vim.split(command, " ")[1]
+---@param command string|string[]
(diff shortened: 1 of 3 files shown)


erroris3:
I have fix all the copilot suggest

nickjvandyke:
Thanks @erroris3 ! I can check this out later today I think.

> Remove polling for port when its not found the opencode to randomly assign a free port and return

I wonder if we should add a special option to `opts.port` for this, like `"random"`? I believe this approach won't find external `opencode`s, so perhaps it should be opt-in (and the default on Windows).

erroris3:
It nice to have but this PR is already done that automatically

erroris3:
I found out that test_port need to be in schedule. It should be fixed now

tonylambert:
I didn't realize someone was working on this. I found a reliable way, I think, to get the CWD on windows with curl against the localhost:<port>/path URL. Check my PR here: https://github.com/NickvanDyke/opencode.nvim/pull/87

nickjvandyke:
Thanks for your preliminary work on this @erroris3 ! And sorry I never got around to it; I didn't have time to address the edge cases. Fortunately seems we've figured that out in #87 ! So I'm gonna close this PR in favor of that one. But thanks again for your help and other contribution 🙂

PR #67, opened 2025-10-29
fix: multiple file references in diagnostics context

chriswritescode-dev:
File referenced multiple times. Potential overflowing context window

### DIFF
--- a/lua/opencode/context.lua
+++ b/lua/opencode/context.lua
@@ -296,7 +296,7 @@ function Context:diagnostics()
     table.insert(
       diagnostic_strings,
       string.format(
-        "%s (%s): %s",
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Thanks for finding this and a fix! While testing, I noticed `opencode` fixed the visual bug when rendering prompts with newlines, so I also updated it to format each diagnostic on its own line 🙂

chriswritescode-dev:
> Thanks for finding this and a fix! While testing, I noticed `opencode` fixed the visual bug when rendering prompts with newlines, so I also updated it to format each diagnostic on its own line 🙂

No problem, I love the plugin. Thank you for creating it! also staying up to date on the latest changes. 😀

PR #69, opened 2025-11-02
feat: add tmux provider

xamcost:
Hi there! This small PR implements a `tmux` provider, enabling users running `neovim` in a tmux session to toggle/start `opencode` in a new pane instead of relying on `snacks.terminal`.

### DIFF
--- a/lua/opencode/provider/tmux.lua
+++ b/lua/opencode/provider/tmux.lua
@@ -14,11 +14,11 @@ Tmux.__index = Tmux
 
 ---Create a new Tmux provider instance
 ---Upon toggling or starting, this will create a new tmux pane to run opencode, in the current window.
----@param opts? opencode.provider.tmux.Opts Configuration options
+---@param opts opencode.provider.tmux.Opts Configuration options
 ---@return opencode.provider.Tmux
 function Tmux.new(opts)
   local self = setmetatable({}, Tmux)
-  self.options = opts and opts.options or "-h"
+  self.options = opts.options
   self.pane_id = nil -- The tmux pane ID where opencode is running
   return self
 end
@@ -91,4 +91,3 @@ function Tmux:show() end
 return {
   Tmux = Tmux,
 }
-


lua/opencode/config.lua
```diff
@@ -168,6 +168,48 @@ local defaults = {
         end
       end,
     },
+    tmux = {
+      split_window_options = "-h", -- Open in a horizontal split by default
```
xamcost: Remark:
1. I am happy to shorten this options' name. I just thought such name makes it self-explanatory.
2. I chose this approach rather than defining a lot of options (such as `pane_size`, `orientation`, ...) to leave more flexibility and avoid complex logic.

lua/opencode/config.lua
```diff
@@ -168,6 +168,48 @@ local defaults = {
...
+        if not vim.env.TMUX then
+          vim.notify(
+            "opencode.nvim: Tmux provider selected but not running inside a tmux session.",
+            vim.log.levels.ERROR,
+            { title = "opencode.nvim" }
+          )
+          return false
+        end
+        return true
+      end,
```
xamcost: Remark: alternatively, we can raise an error. I chose this because it's more user-friendly.

lua/opencode/config.lua
```diff
@@ -168,6 +168,48 @@ local defaults = {
...
+      ---@param self opencode.provider.Tmux
+      toggle = function(self)
+        if not self:check_tmux() then
+          return
+        end
+        -- Check if an `opencode` pane already exists; if so, kill it; if not, split and run
+        local tmux_cmd = string.format(
+          "tmux list-panes -F '#{pane_id} #{pane_current_command}' | grep -q '%s' && tmux list-panes -F '#{pane_id} #{pane_current_command}' | grep '%s' | awk '{print $1}' | xargs -I {} tmux kill-pane -t {} || tmux split-window %s '%s'",
```
xamcost: Important: since this command uses `grep`, `awk` and `xargs` to find the pane where `opencode` is active, it will not work for Windows users... I can try to adapt it in this case ? or put an error message if on Windows and a note in the README. Let me know what you think!

nickjvandyke:
Thanks for this, it looks very thorough at first glance! I'm taking it easy today but will properly review in a day or two 🙂

lua/opencode/config.lua
NickvanDyke: I think an `error()` when this runs on Windows is fine for now. Maybe a comment on `---@class opencode.provider.Tmux` too, so users aren't surprised at runtime. It's a small portion of users and they can still manually manage `opencode`. I have a couple other Windows-specific improvements on the list already anyway haha. But thanks for thinking of them!

xamcost:
Thanks for thoughtful review @NickvanDyke ! I'll work on it tomorrow.

lua/opencode/config.lua
```diff
@@ -168,6 +168,48 @@ local defaults = {
...
+      start = function(self)
+        if not self:check_tmux() then
+          return
+        end
+        -- Check if an `opencode` pane already exists; if not, split and run
+        local tmux_cmd = string.format("tmux split-window %s '%s'", self.split_window_options, self.cmd)
+        vim.fn.system(tmux_cmd)
+      end,
+      ---@param self opencode.provider.Tmux
+      show = function() end, -- users manage tmux panes themselves
```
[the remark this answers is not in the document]
NickvanDyke: Ah I see, thanks for explaining and the link!

That is pretty wild that tmux has three different ways to "show/hide" a pane 😅 would it be problematic for us to "just choose" one? i.e. our method could conflict with how the user likes to manage their tmux panes, and thus feel unintuitive in their workflow?

> I considered initially using the select-pane command in show, but this moves the focus to the opencode pane

Could we easily focus the neovim pane right after? If that's not jank 😆 

Thinking about this more at a high level: if there's no idiomatic and universally-intuitive way to implement `show` for tmux, I agree we should leave it empty. The built-in providers should mold to the underlying tech, not the other way around. Either way it doesn't have to block this PR because it's not critical.

lua/opencode/config.lua
xamcost: > Does `kill`ing a tmux pane actually stop the process, or just the view into it? Just making sure I understand how this `toggle` implementation works. Based off a quick google this seems idiomatic though.

It does kill the process :) 

> I think an `error()` when this runs on Windows is fine for now.

Added in afb0c10f7d89 , and comment in d9a460ad5ecc

xamcost:
Thank you for your insightful review @NickvanDyke ! I applied your suggestions and answered your comments. I ended up also doing a refactoring to create a dedicated `Tmux` class for the `tmux` provider. The rationale behind this was to avoid exposing internal state to users: this class has a `check_tmux` method that isn't intended to me modified by users, and so is the `pane_id` attribute, which keep track of the pane running `opencode`. Of course, I am happy to change/revert this on request.

All in all, this implementation has been done with the idea of sticking to the current behaviour of the plugin. Consequently, all `tmux` operations are done only in the window were the user runs her/his `neovim` instance. If the user runs another pane with `opencode` in another window, it remains undiscovered by the `tmux` provider, so toggling `opencode` will open a new pane with it in that case. 

There is, however, one departure from the `snacks` provider behaviour: when no `opencode` pane are present and `toggle` is used, the focus is shifted towards the newly created `opencode` pane. That's the only case where the focus is shifted. In any other case, the focus remains on the `neovim` pane. It is intentional, I wanted to suggest this behaviour for this particular case, yet if you don't agree, I am happy to change this (it's a one line addition).

Thanks again, and sorry for the verbosity.

choucavalier:
thank you guys :heart:

nickjvandyke:
> I ended up also doing a refactoring to create a dedicated Tmux class for the tmux provider

I like it! The only issue I see is that the type annotations imply the user can pass `provider.tmux.cmd`, toggle, start, and show in their config (which is good), but then `Tmux.new` ignores them (I think). Is it possible to allow overriding those? Given our long discussion about how tmux users have different preferences haha, it'd be nice if they didn't have to completely throw away the built-in tmux provider to customize it a little bit. That's mostly why I structured the snacks provider the way I did. But I agree the class abstraction is nice, so if there's a way to get both then I'm all for it 😁 

> There is, however, one departure from the snacks provider behaviour. when no opencode pane are present and toggle is used, the focus is shifted towards the newly created opencode pane

That makes sense to me! Especially after I recently changed `ask` and `select` to start opencode, it's reasonable to assume that the user wants to use the TUI directly when toggling. I can look into that for the snacks provider too. Should we always focus it in toggle, even when already present, for consistency? (If possible in tmux).

Otherwise this is looking good! It shouldn't break anything, but could you merge `main` to ensure it cooperates with my `ask` and `select` auto-start changes 🙏 

> sorry for the verbosity

It's more than welcome, I'm all about verbosity myself, as you can probably tell 😂 And I'm learning a lot about tmux!

nickjvandyke:
> All in all, this implementation has been done with the idea of sticking to the current behaviour of the plugin. Consequently, all tmux operations are done only in the window were the user runs her/his neovim instance. If the user runs another pane with opencode in another window, it remains undiscovered by the tmux provider, so toggling opencode will open a new pane with it in that case.

I appreciate your eye for consistency 🙂 However feel free to make it work "cross-window" if that's what you think tmuxers would expect. The tmux provider doesn't have to completely align with snacks because they operate differently - snacks runs inside neovim, whereas neovim runs inside tmux.

Overall I thiiiink this still makes sense though. The plugin only calls provider methods if it doesn't find an existing `opencode` (e.g. in another tmux window). So it seems reasonable to me to scope providers to running "alongside" Neovim.

nickjvandyke:
Oh would you mind please adding a check for tmux in `health.lua` too? In its own section, like I did for snacks. It can just have an `ok` "Running inside `tmux`: `"tmux"` provider is available" or similar, and a `warn` for the opposite. Maybe with a unix check too, like you did in the provider. That should save some users a bit of headache 😀

nickjvandyke:
Re the `opencode.provider.Tmux` class, I thought a little more and I think I prefer your approach over the one I used for snacks, so feel free to keep that as-is 🙂 I think I'm over-indexing on users' uncommon desire to customize built-in providers beyond the opts they expose. If there end up being a few different ways to implement a built-in provider's methods, it makes more sense to configure that with an opt than have users override the methods.

xamcost:
Hi @NickvanDyke ! Sorry I've been busy in the last two days! 

> However feel free to make it work "cross-window" if that's what you think tmuxers would expect. The tmux provider doesn't have to completely align with snacks because they operate differently - snacks runs inside neovim, whereas neovim runs inside tmux.

On that aspect, I think the current behaviour of the `snacks` aligns with what `tmux` users would want as well, i.e. discovering `opencode` only in the current window. We're also taking less risk in terms of UX by spawning a new `opencode` instance in the current window rather than shifting the focus to another window in my opinion, because if users use your plugin to access `opencode` from `neovim`, it's likely they want to stay focused on `neovim`.

> Oh would you mind please adding a check for tmux in `health.lua` too?

Sure, will do tomorrow!

> Re the `opencode.provider.Tmux` class, I thought a little more and I think I prefer your approach over the one I used for snacks, so feel free to keep that as-is 🙂

Got it, I will leave it as it is then. In any case, your design allows for users to pass their own provider implementation, so it remains quite flexible.

xamcost:
Hi there! @NickvanDyke Health check added ✅ Let me know what you think! We're ready for another round of review. Let me know if any other changes need to be made. Thank you for your feedback!

nickjvandyke:
Thanks @xamcost ! This is looking great! Once I get some time, I'll update the snacks provider to follow a similar pattern. I think your implementation hides more irrelevant info from the user 😄 

The only thing I wondered is whether `opencode.provider.Opts.tmux`'s type should be a new `opencode.provider.tmux.Opts` class, which has only the `options` field on it? And used as the argument to by `Tmux.new`. That typing would clarify that configuring e.g. `opts.provider.tmux.show` will be ignored, and only `options` has any effect 🙂

nickjvandyke:
Oh and maybe a brief comment on `opencode.provider.Tmux` documenting that it operates in the current tmux window? 🙏

xamcost:
Thank you for your comments @NickvanDyke ! 

> The only thing I wondered is whether `opencode.provider.Opts.tmux`'s type should be a new `opencode.provider.tmux.Opts` class, which has only the `options` field on it? And used as the argument to by `Tmux.new`. That typing would clarify that configuring e.g. `opts.provider.tmux.show` will be ignored, and only `options` has any effect 🙂

Very good idea, thanks. I implemented it in 7a80d344d839 . I took that opportunity to move the `Tmux` class as well as the `tmux.Opts` one in their own file, inside a _provider_ directory. Let me know if you don't like it.

> Oh and maybe a brief comment on opencode.provider.Tmux documenting that it operates in the current tmux window? 🙏

Done as well :)

nickjvandyke:
Perfect, thank you!

Just one super nit remaining: I don't think we have to default `opts.options` inside `Tmux.new`? So that `config.lua#defaults` remains the source of truth 🙂

xamcost:
> Just one super nit remaining: I don't think we have to default `opts.options` inside `Tmux.new`? So that `config.lua#defaults` remains the source of truth 🙂

Indeed, it's better this way. Just done it in 5651ab4c2943

nickjvandyke:
Looks great! Thank you very much for contributing this and iterating on it 😀 I will refactor the snacks provider implementation soon to match!

PR #77, opened 2025-11-16
feat(provider): add kitty provider

thazhemadam:
Add a new provider for kitty terminal emulator.
* The `location`[^1] where the new terminal instance running opencode
should be opened is configurable.
* Allow specifying a password[^2].

[^1]: https://sw.kovidgoyal.net/kitty/launch/#cmdoption-launch-location
[^2]: https://sw.kovidgoyal.net/kitty/remote-control/#cmdoption-kitten-password

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,9 +112,9 @@ If `opencode.nvim` can't find an existing `opencode`, it uses the configured pro
 ```lua
 vim.g.opencode_opts = {
   provider = {
-    enabled = "snacks", -- Default if `snacks.terminal` is available and enabled.
+    enabled = "snacks",
     snacks = {
-      -- Customize `snacks.terminal` to your liking.
+      -- ...
     }
   }
 }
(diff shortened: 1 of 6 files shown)


nickjvandyke:
Ty, I will take a look this week!

nickjvandyke:
I improved the `provider` abstraction after seeing this! See the above commit. Could you please:

1. Use the collapsible format in your README section
2. Implement `provider.name`
3. Implement `provider.stop` - probably using that half of `toggle`
4. Implement `provider.health()` - can just migrate `kitty_check` a bit
5. Add the kitty provider to `provider.lua#list()` and remove it from `config.lua#defaults.provider.enabled` (it iterates through provider healths now)

Should be pretty straightforward, let me know if you have any questions. Then I think this is good to go!

thazhemadam:
@NickvanDyke I've updated the PR with the requested changes. 🙂

nickjvandyke:
I don't use `kitty` so I trust the implementation works, and LGTM otherwise! Thank you 😀

PR #84, opened 2025-11-25
fix: detect and use opencode server even if the comm==bun

### LINKED ISSUE #83: bug: cannot find opencode process

### Operating system/version

NixOS 25.05 (Warbler)

 ============================================================================== opencode:                                                                 1 ⚠️  opencode.nvim ~ - ✅ OK `nvim` version: `0.11.5`. - ✅ OK `opencode.nvim` git commit hash: `fda7eedb597155257817a9f4c9ec17308164657d`. - ✅ OK `vim.g.opencode_opts`: {     events = {       enabled = false     },     provider = {       enabled = "snacks",       snacks = {}     }   }  opencode.nvim [binaries] ~ - ✅ OK `opencode` available with version `1.0.105`. - ✅ OK `curl` available. - ✅ OK `lsof` available: it will be used to auto-find `opencode` if `vim.g.opencode_opts.port` is not set.  opencode.nvim [snacks] ~ - ✅ OK `snacks.input` is enabled: `ask()` will be enhanced. - ✅ OK `snacks.picker` is enabled: `select()` will be enhanced.  opencode.nvim [providers] ~ - ✅ OK Configured `opencode` provider: `snacks`. - ✅ OK The `snacks` provider is available. - ⚠️ WARNING The `tmux` provider is not available — Not running inside a `tmux` session.   - ADVICE:     - Launch Neovim inside a `tmux` session.

### Describe the bug

I upgraded opencode from 1.0.78 to 1.0.105, and opencode.nvim cannot find the instance anymore.
I then updated opencode.nvim to the latest, but the issue persisted.

I think I tracked it down.
```
❯ lsof -w -iTCP -P -n | grep LISTEN
bun       1485060 bart  22u  IPv4 26791345      0t0  TCP 127.0.0.1:34613 (LISTEN)
❯ tr '\0' ' ' </proc/1488140/cmdline
opencode run /nix/store/ymw5140h3x670lwb6l1ifhr3pgi222dx-opencode-1.0.105/lib/opencode/dist/index.js
```
the instance is not started with name "opencode", but "bun", and `find_servers()` does not find it.

### Steps To Reproduce

1. start a new session
2. observe error

### Expected Behavior

stuff works :-)

bartman:
fix #83

- calling out to `lsof` now allows matches on `opencode` and `bun` in `comm` field
- after first pass, make sure the command line actually contains `opencode` string

### DIFF
--- a/lua/opencode/cli/server.lua
+++ b/lua/opencode/cli/server.lua
@@ -24,23 +24,13 @@ end
 ---@param pid number
 ---@param string string
 ---@return boolean
-local function check_proc_cmdline_for_string(pid, string)
-  local cmdline_path = '/proc/' .. tostring(pid) .. '/cmdline'
-
-  -- Check if the file exists and is readable
-  if vim.fn.filereadable(cmdline_path) == 1 then
-    -- Read the file (it's a single line with null separators)
-    local cmdline = vim.fn.readfile(cmdline_path)[1]
-
-    -- Check if string appears anywhere in the command line
-    if string.find(cmdline, string) then
-      return true  -- Found it
-    else
-      return false  -- Not found
-    end
-  else
-    return false  -- File doesn't exist (invalid PID or process gone)
+local function check_proc_cmdline_for_string(pid, needle)
+  local output = exec("ps -p " .. pid .. " -ww -o args=")
+  if vim.v.shell_error ~= 0 then
+    return false
   end
+  local result = string.find(output, needle, 1, true)
+  return result ~= nil
 end
 
 -- derive the CWD from a PID


nickjvandyke:
Thanks! I don't think `/proc` supports MacOS unfortunately - is there any way to do this without?

bartman:
> Thanks! I don't think `/proc` supports MacOS unfortunately - is there any way to do this without?

Now I understand why you were using `lsof` to get the `cwd`... I used `/proc/` to avoid the exec step... but that's a regression.

I'll rework it to use lsof.

bartman:
@NickvanDyke can you check on OSX if this returns the full command line?

(substitute PID with your opencode PID)

```
❯ ps -p $PID -ww -o args=
opencode run /nix/store/ymw5140h3x670lwb6l1ifhr3pgi222dx-opencode-1.0.105/lib/opencode/dist/index.js
```

bartman:
I've rewritten the code to use `lsof` and `ps` executables, instead of poking at `/proc`.
Hopefully that works for you as well.

Do you think I should add `ps` to the list of checks in `lua/opencode/health.lua` ?
Or is that standard on OSX ?

nickjvandyke:
Thank you!

`ps` is available standard on MacOS. iirc based on a comment above the `lsof` usage though, its output changed a bit by platform. Hence I experimented with `lsof` flags until it completely replaced `ps`. However I'm not sure `lsof` can accomplish what you did with `ps` here: checking command args too 🤔 

I'm trying to stay away from coding during the holidays haha but I'll give this a try after 🙂 Let me know if the issue persists for you on newer versions of `opencode` too. Or I wonder if it's worth asking them to keep the command name stable, since it's essentially part of the interface.

Mellowchan:
I've tested this PR and I can confirm it works on NixOS with opencode v1.0.114

nickjvandyke:
I wonder if this is a bit platform-specific? `ps -e | grep opencode` still doesn't return `bun ...` for the `opencode` process for me - just `opencode` as always. We should still fix it for other platforms (or ask `opencode` what changed on their end), just making sure we know what's what.

nickjvandyke:
Or I wonder - did you install `opencode` via `bun`? I installed via MacOS homebrew and don't even have `bun` on my system. In which case we may also want to check `npm` and `paru` installs.

Mellowchan:
> I wonder if this is a bit platform-specific? `ps -e | grep opencode` still doesn't return `bun ...` for the `opencode` process for me - just `opencode` as always. We should still fix it for other platforms (or ask `opencode` what changed on their end), just making sure we know what's what.

I assume it's just decision of the package maintainer to use `bun` instead of `npm`. It's not platform specific since `bun` can be used on MacOS too, and other Linux package managers might also use `bun` instead of `npm`, mainly due to performance and resource usage benefits.

> Or I wonder - did you install `opencode` via `bun`? I installed via MacOS homebrew and don't even have `bun` on my system. In which case we may also want to check `npm` and `paru` installs.

On NixOS `opencode` <a href="https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/op/opencode/package.nix#L45">gets build like this</a>. Previously `opencode` used `npm` on Nixos, but in <a href="https://github.com/NixOS/nixpkgs/commit/27d49dfeb41c3adc279254eef671fa641c1c48f7">this commit</a> the maintainer switched to `bun`.

Isn't `paru` just an extension for AUR and not a Node.js runtime, if I'm not mistaken?

nickjvandyke:
I merged #90 because`pgrep` handled edge cases a little better, but thank you @bartman for reporting this issue with a preliminary fix! 😀

PR #87, opened 2025-11-28
feat(server): detect the right opencode CWD for running instances on Windows

tonylambert:
Leveraging a powershell script and curl, extract the working directory of each running opencode instance and return a list of servers similar to the `lsof` implementation on unixes.

I have not tested this rigorously, but I can successfully run multiple nvim instances in different directories and properly send information from nvim to the correct opencode instance using these changes.

There may be a quicker way to get the pid and ports, though. I was unable to find any way to get the working directory for the running opencode instances *without* making the curl "call" to the localhost:<port>/path endpoint. Because get_port operates on a promise model, though, I didn't see negative startup impact.

### DIFF
--- a/lua/opencode/cli/server.lua
+++ b/lua/opencode/cli/server.lua
@@ -120,6 +120,7 @@ local function populate_cwd(process)
   assert(vim.fn.executable("curl") == 1, "`curl` executable not found")
 
   -- Query each port synchronously for working directory
+  -- TODO: Migrate `client.lua` to use `vim.system` and move this there.
   local curl_result = vim
     .system({
       "curl",


nickjvandyke:
This is great, thank you! I didn't know about `/path`. Wdyt about refactoring the Unix path to use that too for CWD? i.e. we use platform-specific powershell or `lsof` to get `opencode` ports, then `curl /path` on all platforms. Less branches and I bet in some weird edge case (now or in the future), `/path` is more reliable. I'm happy to test the Unix path if you're willing to roughly refactor that, otherwise I can take care of it later after merge.

Feel free to do away with or simplify the `Server` type if needed; it's not really that useful haha.

tonylambert:
Makes perfect sense. I wanted to make a small change to see where your head was at first. I'm on my linux machine, so I haven't tested it on Windows yet, but take a look at https://github.com/NickvanDyke/opencode.nvim/pull/87/commits/9aa9f1c86e5ae3ce428198f2b3201d9b9fcf34c7 and let me know if this is on the right track for what you are looking for.

tonylambert:
I did some testing by remoting into my Windows machine, and it is working as expected.

nickjvandyke:
Ty! That's looking good. I'm thinking we can also remove this:

```lua
elseif use_lsof then
      -- Fall back to lsof on Unix systems if available
      local lsof_output = exec("lsof -w -a -p " .. proc.pid .. " -d cwd")
      cwd = lsof_output:match("%s+(/.*)$")
```

And always go with the `use_curl` path? `curl` is already a required dependency to actually interact with `opencode` so we don't need an `lsof` fallback here.

tonylambert:
Update to remove the `lsof` fallback. Let me know your thoughts.

tonylambert:
FYI, @NickvanDyke , I ran `stylua` on client.lua and server.lua and fixed the lua_ls complaint in client.lua. I pushed two commits for these. The checks should pass now, I think.

nickjvandyke:
Ty @tonylambert ! I relaxed the workflow approval settings just now, but if that doesn't apply to existing PRs, you can open a PR to `main` in your fork to run + verify them if needed.

This PR and https://github.com/NickvanDyke/opencode.nvim/pull/90 will have some merge conflicts. Do you mind if I merge that first? I think it affects more users and resolving conflicts on this PR seems easier than the inverse.

tonylambert:
No concerns. You are in charge :)

nickjvandyke:
I see rn it normalizes paths to backward slashes. Wdyt about normalizing to forward slashes instead, if that works on Windows? And maybe skip the normalization/replacement entirely on non-Windows? It's unlikely but I would guess that'd minimize the number of users potentially affected by some unforeseen filepath edge case.

tonylambert:
Resolved the pathing feedback. Only changing the paths on Windows, now.

nickjvandyke:
Thank you! I merged https://github.com/NickvanDyke/opencode.nvim/pull/90 - once we resolve conflicts I think this is good to go. Note it now uses `pgrep` on unix instead of `lsof` after that PR 🙂

tonylambert:
I need to test on Windows yet, but I did merge my changes with the merged in pgrep solution. However, the pgrep solution still uses lsof in order to get the ports for the opencode processes. So, right now, the way I did it right now is on unixes it still does "pgrep -> lsof" boom, your server list and on windows it does "powershell script (pid, ports) -> curl (cwd)." On unixes it could to "pgrep (pid) -> lsof (port) -> curl (cwd)" if you'd like. Sadly, since pgrep isn't surfacing a port, lsof still needs to stick around?

tonylambert:
Verified that it functions on Windows.

nickjvandyke:
> On unixes it could to "pgrep (pid) -> lsof (port) -> curl (cwd)" if you'd like

yeah that was my intention, sorry I should have clarified!

While `lsof` for CWD works atm, I worry that in the future `opencode` could add the ability to change its CWD internally. The process's CWD (reflected in `lsof`) may or may not reflect that. But we know that `/path` would.

lua/opencode/cli/server.lua
```diff
@@ -30,13 +41,6 @@ local function find_servers()
     )
   end
 
-  if vim.fn.executable("pgrep") == 0 then
-    error(
-      "`pgrep` executable not found in `PATH` to auto-find `opencode` — please install it or set `vim.g.opencode_opts.port`",
-      0
-    )
-  end
-
```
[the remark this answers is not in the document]
tonylambert: Agreed. Thank you for catching that. I merged your changes in poorly. 🤦🏻

tonylambert:
I did some light testing on Linux, Macos, and Windows. They are all working as expected. I will say that the slow startup of powershell makes the Windows implementation seem a bit laggy. I'll look around for a different solution (read, non-powershell solution).

In the meantime, it works on Windows and if anyone is running under WSL, it wouldn't use the powershell version. I don't have any kind of numbers to back this up, but I feel confident saying that a fair number of the "Windows" users of nvim are running it under WSL.

I believe I have addressed the remaining concerns after the two recent commits. Please advise.

Thanks!

lua/opencode/cli/server.lua
```diff
@@ -20,19 +31,21 @@ local function exec(command)
   return output
 end
 
----@return Server[]
-local function find_servers()
-  if vim.fn.executable("lsof") == 0 then
-    -- lsof is a common utility to list open files and ports, but not always available by default.
```
jerome-benoit: oft ?

lua/opencode/cli/server.lua
```diff
@@ -43,53 +56,142 @@ local function find_servers()
...
+  return processes
+end
+
+-- Find the working directories of running instances of opencode.
+---@param processes {pid: number, port: number}[]
+---@return Server table
+local function populateWorkingDirectories(processes)
+  local servers = {}
+
+  if vim.fn.executable("curl") == 0 then
```
jerome-benoit: Making curl a transversal required tool is debatable: 
- can't a simple and portable LUA http client be used instead? 
- if no, can't we have a fallback on well known HTTP client CLI such as wget? or even lsof usage on *nix?

lua/opencode/cli/server.lua
```diff
@@ -20,19 +31,21 @@ local function exec(command)
...
+  if vim.fn.executable("pgrep") == 0 then
     error(
-      "`lsof` executable not found in `PATH` to auto-find `opencode` — please install it or set `vim.g.opencode_opts.port`",
+      "`pgrep` executable not found in `PATH` to auto-find `opencode` — please install it or set `vim.g.opencode_opts.port`",
       0
     )
   end
 
-  if vim.fn.executable("pgrep") == 0 then
+  -- On Unix, use lsof to find cwd of processes
```
jerome-benoit:
```suggestion
```

lua/opencode/cli/server.lua
tonylambert: @NickvanDyke - Do you have a viewpoint on this?

nickjvandyke:
I think this is all set! Thank you for this great solution and iterating and testing so much! 😀 

If anyone feels inspired, please feel free to explore non-`curl` solutions separately! 🙂

PR #89, opened 2025-11-28
fix: change handle_sse and handle_json to properly parse streaming vim.fn.jobstart on_stdout calls

tghosgor:
This PR changes handle_sse and handle_json functions to correctly handle `vim.fn.jobstart`'s `on_stdout` calls as described [here](https://neovim.io/doc/user/channel.html#channel-lines). See individual commits to see what they do to fix the issues.

### DIFF
--- a/lua/opencode/cli/client.lua
+++ b/lua/opencode/cli/client.lua
@@ -38,19 +38,25 @@ local function handle_sse(data)
   return responses
 end
 
+local json_state = {
+  buffer = {},
+}
+
 ---@param data table
 ---@return table
 local function handle_json(data)
-  for _, line in ipairs(data) do
-    if line == "" then
-      return {}
-    end
-    local ok, response = pcall(vim.fn.json_decode, line)
+  if #data == 1 and data[1] == "" then -- this is eof
+    local full_data = table.concat(json_state.buffer)
+    json_state.buffer = {}
+
+    local ok, response = pcall(vim.fn.json_decode, full_data)
     if ok then
       return { response }
     else
-      vim.notify("JSON decode error: " .. line, vim.log.levels.ERROR, { title = "opencode" })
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Seems stable to me! Could you just fix the tiny issues with the CI checks please?

Thank you very much for tackling this, it's bothered me for a while but I didn't have the time it needed 😀

tghosgor:
> Seems stable to me! Could you just fix the tiny issues with the CI checks please?

I think I've fixed them now.

> 
> Thank you very much for tackling this, it's bothered me for a while but I didn't have the time it needed 😀

Yeah me too. It just kept happening way too much on a repository I'm working on so it had to go away. :smile:

nickjvandyke:
I think this is the typing to make LuaLS happy 😀 

```lua
      local responses
      if is_sse then
        responses = handle_sse(data)
      else
        responses = handle_json(data)
      end
      if responses and callback then
        for _, response in ipairs(responses) do
          callback(response)
        end
      end
```

tghosgor:
> I think this is the typing to make LuaLS happy 😀
> 
> ```lua
>       local responses
>       if is_sse then
>         responses = handle_sse(data)
>       else
>         responses = handle_json(data)
>       end
>       if responses and callback then
>         for _, response in ipairs(responses) do
>           callback(response)
>         end
>       end
> ```

I am not sure even that's enough. I'll fix my lua lsp in a sec.

tghosgor:
Huh I didn't know I could run these "actions" in my fork. I am not that familiar with modern features of github, sorry :).

It turned out to be about the annotations on `handle_*` functions. It should be fine now.

nickjvandyke:
No problem, that's news to me too 😂 

Given the issue's infrequent nature, this will need some time in the wild to verify. But it seems correct to me, and at the very least, didn't break anything. So let's go for it!

Thanks again! 🙂

PR #91, opened 2025-11-30
Add diff viewing

chuan2984:
## Summary
Add PR-style diff review UI for `opencode` changes.

This is my rough draft based on some discussion [here](https://github.com/NickvanDyke/opencode.nvim/discussions/76). I'll be honest and that the `session_diff.lua` was mostly AI coded, but was verified and checked for accuracy.

Please pull it if you are interested, its been working well for me so far

## Screenshots
image

## Key Points

- **SSE**: Uses `message.updated` event (contains only current cycle diffs, unlike `session.diff` which 
accumulates everything) for capturing the `before`, and `after` for only files touched in the current cycle.
- **Diff display**: 
   - Simple `vim.diff()` unified view
   - A more complex `pr review` style enhanced diff view that works with `hunk staging`

**note:**  `DiffView.nvim` sadly does not really support what we want, since it only works with git rev not local temp files(we are not using `git`, we dont get much from `DiffView.nvim` other than the nice UI

<details><summary>Screenshot of enhanced diff</summary>
image
</details> 

 
## Files
- `lua/opencode/diff.lua` - Core review logic
- `plugin/events/session_diff.lua` - Event listener
- `DIFF_FEATURE.md` - Full documentation

## TODOs
- [ ] better abstraction/file organization
- [ ] better user command
- [ ] the ability to see the most recent diff
- [ ] slightly better-looking UI
- [ ] the ability to set keymaps
- [ ] `plan` mode should show diff but non-editable
- [ ] What should happen to diff if files outside of current dir are modified
- [ ] Add toggle file panel keymap
- [ ] Delay the events until idle or that do it without setting focus?
- [ ] MORE testing~~~

## Bugs
- [ ] The highlighting on the current file can be off in certain scenarios, treesitter problem. Incorrect timing, re-open fixes it
- [ ] Shows only new addition for my lua files, no idea why, this is because of symlink, This is a opencode problem. 

## Nice to have?
- any permission event dont automatically switch the focus back to nvim

### DIFF
--- a/lua/opencode/diff.lua
+++ b/lua/opencode/diff.lua
@@ -408,30 +408,35 @@ function M.enhanced_diff_show_file(index)
   -- Open the before buffer on the left
   vim.api.nvim_set_current_buf(before_buf)
 
-  -- Force reload if buffer already loaded to ensure FileType event fires
-  -- This ensures Treesitter highlighting attaches properly
-  local buf_exists = vim.fn.bufexists(file_entry.actual_file) == 1
-  if buf_exists then
-    local existing_buf = vim.fn.bufnr(file_entry.actual_file)
-    if vim.api.nvim_buf_is_loaded(existing_buf) then
-      vim.api.nvim_buf_call(existing_buf, function()
-        vim.cmd("edit!")
-      end)
-    end
-  end
-
   -- Open the actual file (after) on the right
   vim.cmd("rightbelow vertical diffsplit " .. vim.fn.fnameescape(file_entry.actual_file))
+
+  -- Store window references (do this before enabling diff mode)
+  M.state.enhanced_diff_left_win = vim.fn.win_getid(vim.fn.winnr("h"))
+  M.state.enhanced_diff_right_win = vim.fn.win_getid()
+
+  -- Ensure filetype is set and Treesitter attaches to the right buffer
+  local after_buf = vim.api.nvim_win_get_buf(M.state.enhanced_diff_right_win)
+  vim.api.nvim_buf_call(after_buf, function()
+    -- Force filetype detection
+    vim.cmd("filetype detect")
+    
+    -- Explicitly start Treesitter if available
+    local ft = vim.bo[after_buf].filetype
+    if ft and ft ~= "" then
+      local has_ts, ts_start = pcall(require, "vim.treesitter")
+      if has_ts then
+        local lang = vim.treesitter.language.get_lang(ft) or ft
+        pcall(vim.treesitter.start, after_buf, lang)
+      end
+    end
+  end)
 
   -- Enable diff mode
   vim.cmd("wincmd p")
   vim.cmd("diffthis")
   vim.cmd("wincmd p")
   vim.cmd("diffthis")
-
-  -- Store window references
-  M.state.enhanced_diff_left_win = vim.fn.win_getid(vim.fn.winnr("h"))
-  M.state.enhanced_diff_right_win = vim.fn.win_getid()
 
   -- Set up keybindings for both diff windows
   local keymap_opts = { buffer = true, nowait = true, silent = true }


nickjvandyke:
> sometimes closing neovim would not close the manage opencode session

I didn't test this super extensively, especially with new providers. lmk if you find a cause or fix along the way!

froats:
Hi, how is this going?

chuan2984:
@froats , still in testing mode, i havent had too much chance to work on this. Theres something wrong with opening the current version, treesitter sometimes is not correctly attached. Other than that and some QoL stuff, i have been using it without a problem. Although i dont really use the diff features(accept hunk/reject hunk)that often(hunk definition is not that accurate). Please give it a try if you are interested. It is 80% there.

I also just attached a screenshot as well.

naowalrahman:
Opencode actually stores snapshots (git commits in a shadow repo) per-message in `~/.local/share/opencode/snapshot`, so I wonder if you could use something like [codediff.nvim](https://github.com/esmuellert/codediff.nvim) plus a picker to let users pick the messages they want to diff between for the currently open session? (or even simpler, a session diff and last message diff).

chuan2984:
> Opencode actually stores snapshots (git commits in a shadow repo) per-message in `~/.local/share/opencode/snapshot`, so I wonder if you could use something like [codediff.nvim](https://github.com/esmuellert/codediff.nvim) plus a picker to let users pick the messages they want to diff between for the currently open session? (or even simpler, a session diff and last message diff).

thank you, didnt know about this. Will check it out tomorrow on MLK day.

naowalrahman:
> > Opencode actually stores snapshots (git commits in a shadow repo) per-message in `~/.local/share/opencode/snapshot`, so I wonder if you could use something like [codediff.nvim](https://github.com/esmuellert/codediff.nvim) plus a picker to let users pick the messages they want to diff between for the currently open session? (or even simpler, a session diff and last message diff).
> 
> thank you, didnt know about this. Will check it out tomorrow on MLK day.

I made a PR #137!

nickjvandyke:
Implemented with Neovim's `:diffpatch` for consistency and maintainability. Thanks for taking a crack at this though!

chuan2984:
> Implemented with Neovim's `:diffpatch` for consistency and maintainability. Thanks for taking a crack at this though!

I apologize for not going through with it. It has always bothered me to leave the work in this state. Im glad you took it over. 

Im trying to test it out, but I dont see anything when opencode is making edits to files, I still only see the diff inside opencode and not inside a neovim tab. Am I doing something wrong?

I believe its the SSE thats not categorized as `edit`? This started to happen with my branch when not using `ask`, a normal linked server wouldn't trigger the diff thing I was building

nickjvandyke:
All good!

> This started to happen with my branch when not using ask, a normal linked server wouldn't trigger the diff thing I was building

Try opening ask or select (and then closing) to connect to the server first. Otherwise the plugin won't listen to its SSEs. If I understand your flow correctly.

chuan2984:
> Try opening ask or select (and then closing) to connect to the server first. Otherwise the plugin won't listen to its SSEs. If I understand your flow correctly.

Im still not getting any new tabs. What I did, open neovim, did ask @this `add a comment on top of this file` 
image

michaelrommel:
I am also just trying this out and I never see a diff window. Regardless, if opencode is running in a separate terminal windows and the plugin is connecting to it or if it is running in a neovim terminal split. I set the permission for edit to ask and I do get a requester popup for once, always, reject, but no diff window.

That is what I see, and as soon as I select "once" the edit appears.

image

michaelrommel:
Sorry, the version with the patch was not published. I needed to pin the commit in `Lazy` to get this version and now it works. Thanks for the **awesome** plugin!

chuan2984:
> Sorry, the version with the patch was not published. I needed to pin the commit in `Lazy` to get this version and now it works. Thanks for the **awesome** plugin!

Im on `main` and im not seeing this, do i have to set the permission for edit to ask everytime?

michaelrommel:
I just had to enable this in `opencode.json`:

image

chuan2984:
im on commit 8804ffb, the permission window is not brought up in neovim either.

michaelrommel:
For me the original neovim tab looks like this:

image

and the tab to which neovim gets switched looks like this:

image

I can then accept or reject the diff. I do not understand, however, how one could selectively accept several hunks. That does not seem possible, either all, one or none.

chuan2984:
Im still trying to get this to work, i copied the SSE example into my config and i can see all the events coming in, and when I just entered neovim I can select a server to connect to, but I still do not get the diff tab..hmmm... ill look into this a bit later

michaelrommel:
But I think it does not work correctly. I have asked as an example to amend all pub functions with a doc comment. The first three iterations, the diff was shown correctly and although I accepted each one, after the fourth time, the diff showed up, but no lines were shown as diff. The whole file was collapsed in a fold and the line suggested as edit did not show up at all. I would rather stick to the method before and accept the diffs, via a popup or in the TUI...

chuan2984:
I got this to work, it really needs to ask for permission to edit each time. For some reasons when I quit neovim with server open, the server just dangles in the background. I needed to `pkill -9 opencode` to have neovim automatically hook into the current server session. Then I was able to see the diff, but as soon as I hit 'always', the diff window no longer appears, which i guess is expected. 

I need to give this workflow a try for a few days. The way that I configured the diff viewer(in this PR) was done after the fact, the `hunk reject` and `hunk accept` were basically `revert` and `do nothing` respectively. Since the edits were not stopped, it just does a before and after comparison of the snapshot taken before the prompt started. This allowed me to see and understand the whole picture before I made any further modifications, and it was rarely used as a way to edit but to understand what just happened.

michaelrommel:
I think I will let opencode just change the files. I have the git actions set to ask. So I think i will just use my git plugin and in that one then collect the hunks I want to accept into a commit after opencode finishes all changes. This is much easier to get an overview on what was changed in total, especially since on larger changes, the individual diffs do not reflect any coherent state.

nickjvandyke:
> For some reasons when I quit neovim with server open, the server just dangles in the background

`opencode` bug unfortunately - see #118 

> do i have to set the permission for edit to ask everytime?

Ah sorry I was confused. Yes, the plugin relies on the diff sent in the `edit` permission request.

> the diff showed up, but no lines were shown as diff

Yeah, this is an occasional bug. I'm still not sure why. I think the sent diff might be malformed, at least for what Neovim expects.

nickjvandyke:
> I would rather stick to the method before and accept the diffs, via a popup or in the TUI...

@michaelrommel pushed a change just now to continue using the generic permissions popup for edits when `opts.events.permissions.edits.enabled == false` 🙂

michaelrommel:
> the diff showed up, but no lines were shown as diff
> 
> Yeah, this is an occasional bug. I'm still not sure why. I think the sent diff might be malformed, at least for what Neovim expects.

I think it is pretty difficult to manage those integrations, especially since LLMs have no "strict, reproducible" behaviour... But a great job, that you did!

chuan2984:
@nickjvandyke good day, I have been trying with this option, I cant say this is what I would find useful. The prompt to check before each `edit` is quite useless since I do not have a grasp on the bigger picture. It is also buggy in my experience since the window doesnt always popup(i think this has to do with not attaching to the right opencode instance) 

I wanted to get your opinion on this once again before I revive the PR(I will set a timeline to finish it if I do). Do you think a `after-the-fact` kind of diff viewing belongs to this project(using built-in vim.diff), basically what the branch was trying to do. 
I really dont think people are still reading the generated code line by line, at least not when its generated. If anything, having a unified view of what happened after multiple agents are done would be the future imo.

naowalrahman:
Personally I'd love an after-the-fact diff! I had a janky attempt in #137 but it was unnecessarily reverse engineering opencode snapshots. 

Opencode actually exposes a [`/session/:id/diff`](https://opencode.ai/docs/server/#:~:text=GET-,/session/%3Aid/diff,-Get%20the%20diff) endpoint so a per-session diff shouldn't be hard to implement.

chuan2984:
> Opencode actually exposes a [`/session/:id/diff`](https://opencode.ai/docs/server/#:~:text=GET-,/session/%3Aid/diff,-Get%20the%20diff) endpoint so a per-session diff shouldn't be hard to implement.

nice, I remember looking at your PR. I was also replicating this by storing a snapshot beforehand and comparing it to the current code after every `update` event.
I dont mind making a separate plugin that is just dedicated to supplementing the viewing part either if it is eventually going to be duplicated by `opencode`

nickjvandyke:
Thanks for asking first. I think viewing the session diff might duplicate other git plugins too closely. I can't speak for everyone's workflow, but I'd think a new session usually warrants a commit after the previous session. Thus the session diff equals the entire diff since the last commit, which is easily interacted with by other git plugins/tools.

> Do you think a after-the-fact kind of diff viewing belongs to this project(using built-in vim.diff)

I'm usually happy to expose such integrations for power users. i.e. make it easier for them to hook their own diff plugin up, even if we don't provide a diff ourselves. That kind of thing has a good value to maintenance burden ratio. Perhaps additionally defaulting to nvim's built-in diff tools like I did with the edit permission. I'm curious how nvim's diff would handle accepting/rejecting changes here - would it "just work", or need special handling to reconcile the files and the session diff.

So, at the least we can easily add a server function that calls the session diff endpoint. Then users can call that and pass the diff to their tool of choice. Maaaaaybe also adding a `select()` entry to view the session diff in nvim's diff tool - and customize the backend - depending on how complex that code is.

naowalrahman:
I think exposing a function calling the diff endpoint is sufficient -- Neovim's current diff tool supports just files and directories, so the plugin would have to create a 2 temporary directories containing the before and after files returned by the endpoint, which can get kind of nasty and is probably better off as a "plugin to the plugin" haha

Issue #92, opened 2025-12-04
bug: wezterm toggle opening new panes instead of hiding the visible one


tonylambert:
### Did you check docs and existing issues?

- [x] I have read all the plugin docs
- [x] I have searched the existing issues
- [x] I have searched the existing [opencode issues](https://github.com/sst/opencode/issues)

### Operating system/version

Linux

### `:checkhealth opencode` output

============================================================================== opencode:                                                                 4 ⚠️  opencode.nvim ~ - ✅ OK `nvim` version: `0.11.5+v0.11.5`. - ✅ OK `opencode.nvim` git commit hash: `d9b147448a4c713bee4f70e11d5670796be958ff`. - ✅ OK `vim.g.opencode_opts`: {     provider = {       enabled = "wezterm",       wezterm = {         direction = "right",         percent = 35,         top_level = false       }     }   }  opencode.nvim [binaries] ~ - ✅ OK `opencode` available with version `1.0.128`. - ✅ OK `curl` available. - ✅ OK `lsof` available: it will be used to auto-find `opencode` if `vim.g.opencode_opts.port` is not set.  opencode.nvim [snacks] ~ - ⚠️ WARNING `snacks.nvim` is not available: `ask()` and `select()` will not be enhanced.  opencode.nvim [providers] ~ - ✅ OK Configured `opencode` provider: `wezterm`. - ⚠️ WARNING The `snacks` provider is not available — `snacks.nvim` is not available.   - ADVICE:     - Install `snacks.nvim` and enable `snacks.terminal.` - ⚠️ WARNING The `kitty` provider is not available — KITTY_LISTEN_ON environment variable is not set.   - ADVICE:     - Enable remote control in `kitty`. - ✅ OK The `wezterm` provider is available. - ⚠️ WARNING The `tmux` provider is not available — `tmux` executable not found in `$PATH`.   - ADVICE:     - Install `tmux` and ensure it's in your `$PATH`.

### Describe the bug

The toggle command running in wezterm with the wezterm provider configured is launching new panes every time toggle is called instead of showing/hiding (or creating/destroying) with subsequent calls to toggle. Steps to reproduce:

I am going to do some more digging. I concede that it may be something about my setup, but I'll report it for now and see 

### Steps To Reproduce

<Steps assume a keybind that runs the function: require("opencode").toggle()>

1. Open an instance of wezterm.
2. Open neovim in said instance of wezterm.
[...]

PR #93, opened 2025-12-04
Fix pane spawning bug

chuan2984:
Switching to pane lookup by the title `opencode` ended up breaking the behavior, because the pane title is actually determined by the foreground process—`node` in this case—and WezTerm doesn’t provide a way to override it. This change also removes the unnecessary dependency on `jq`, which was an oversight.

There were two alternative approaches that I decided not to take:
1. Matching by `cwd` and the `title` `node`. This can work, but it’s still not fully deterministic.
2. Using a wrapper script that runs `opencode` under a process name of `opencode`. This would require extra setup from users and feels too hacky.

Fixes https://github.com/NickvanDyke/opencode.nvim/issues/92

### DIFF
--- a/lua/opencode/provider/wezterm.lua
+++ b/lua/opencode/provider/wezterm.lua
@@ -52,7 +52,10 @@ function Wezterm.health()
   return true
 end
 
----Get the `wezterm` pane ID where `opencode` is running.
+---Retrieve the `wezterm` pane ID associated with the running `opencode` instance.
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Thanks for fixing this! I think tracking only panes spawned by the provider is okay, and might even make more sense. The plugin will never call the provider when it finds a manually-started `opencode` anyway. I'm doing some `tmux` troubleshooting and encountered a similar issue. So I think we can standardize this approach across providers.

PR #98, opened 2025-12-19
fix(server): skip invalid processes when discovering opencode servers

### LINKED ISSUE #99: bug: Failed to get working directory for opencode

### Operating system/version

MacOS 26.1

opencode:                                                                 2 ⚠️  opencode.nvim ~ - ✅ OK `nvim` version: `0.11.4`. - ✅ OK `opencode.nvim` git commit hash: `bc42d98c8efb23022d32fde878b0c07018eb73a8`. - ✅ OK `vim.g.opencode_opts`: {}  opencode.nvim [binaries] ~ - ✅ OK `opencode` available with version `1.0.169`. - ✅ OK `curl` available. - ✅ OK `pgrep` available: it will be used to auto-find `opencode` if `vim.g.opencode_opts.port` is not set. - ✅ OK `lsof` available: it will be used to auto-find `opencode` if `vim.g.opencode_opts.port` is not set.  opencode.nvim [snacks] ~ - ✅ OK `snacks.input` is enabled: `ask()` will be enhanced. - ✅ OK `blink.cmp` is available: `opts.ask.blink_cmp_sources` will be registered in `ask()`. - ✅ OK `snacks.picker` is enabled: `select()` will be enhanced.  opencode.nvim [providers] ~ - ✅ OK Configured `opencode` provider: `snacks`. - ✅ OK The `snacks` provider is available. - ⚠️ WARNING The `kitty` provider is not available — KITTY_LISTEN_ON environment variable is not set.   - ADVICE:     - Enable remote control in `kitty`. - ⚠️ WARNING The `wezterm` provider is not available — `wezterm` executable not found in `$PATH`.   - ADVICE:     - Install `wezterm` and ensure it's in your `$PATH`. - ✅ OK The `tmux` provider is available. - ✅ OK The `terminal` provider is available.

### Describe the bug

When I try to open the `Ask opencode` (or any other command), it opens the right side panel, but raises the following error and does not recognizes my cwd:

```
Failed to get working directory for opencode process: 55472
```

Current setup:
- Opencode via homebrew (v. 1.0.169)
- Lazyvim
- Kitty
- Tmux

### Steps To Reproduce

1. Open opencode
2. The pane on the right opens
3. The error happens
4. If you close and try to open again, it returns only the error, not the panel anymore.

Image

### Expected Behavior

It's expected to open the small input saying "Ask opencode"

juanperi:
## Summary
- Fix server discovery failing when `pgrep -f 'opencode'` matches non-opencode processes (e.g., language servers started by opencode like elixir-ls at `~/.local/share/opencode/bin/elixir-ls-master/`)
## Problem
When discovering opencode servers, `find_servers()` uses `pgrep -f 'opencode'` which can match processes that have "opencode" in their path or arguments but aren't actual opencode servers. For example, the Elixir language server spawned by opencode:
```
beam.smp ... -extra /Users/jperi/.local/share/opencode/bin/elixir-ls-master/release/launch.exs
```
Previously, if any matched process failed the `/path` endpoint check, `find_servers()` would throw an error and abort entirely, preventing the plugin from finding valid opencode servers.
## Solution
Wrap `populate_cwd()` in `pcall()` and filter out processes that fail validation, rather than erroring out on the first failure. This allows the plugin to skip invalid processes and continue searching for valid opencode servers.

This patch has been working on my local setup, but I'm not sure if it's the "right" fix, so inputs welcome!

Maybe fixes #99

### DIFF
--- a/lua/opencode/cli/server.lua
+++ b/lua/opencode/cli/server.lua
@@ -161,8 +161,20 @@ local function find_servers()
     error("No `opencode` processes found", 0)
   end
 
+  -- Filter out processes that aren't valid opencode servers.
+  -- pgrep -f 'opencode' may match other processes (e.g., language servers
+  -- started by opencode) that have 'opencode' in their path or arguments.
   ---@type opencode.cli.server.Server[]
-  local servers = vim.tbl_map(populate_cwd, processes)
+  local servers = {}
+  for _, process in ipairs(processes) do
+    local ok, server = pcall(populate_cwd, process)
+    if ok then
+      table.insert(servers, server)
+    end
+  end
+  if #servers == 0 then
+    error("No valid `opencode` servers found", 0)
+  end
   return servers
 end
 


nickjvandyke (commented):
Nice, thanks! I must have misunderstood how pgrep works when we migrated to it. I didn't intend for a failure here to fail the whole thing either, so I guess that was a mistake on my part. I don't think you need to error on zero servers - the caller does that already iirc. Otherwise LGTM!

nickjvandyke:
Actually I guess that error is consistent with the other code, nevermind!

PR #103, opened 2025-12-23
Suppress JSON decode errors during rapid scrolling

timseriakov:
## Problem
When rapidly scrolling or sending multiple commands in quick succession (e.g., using `Ctrl-U`/`Ctrl-D` in a terminal wrapper), the plugin occasionally receives incomplete or empty SSE/JSON chunks. This triggers a `vim.notify` with an error level, which can be disruptive to user experience.

## Solution
1. Add empty data checks before `json_decode` calls in both SSE and JSON handlers
2. Change error notification level from `ERROR` to `DEBUG` 
3. Errors only shown when `vim.g.opencode_debug` is enabled (default: `false`)

## Changes
Modified `lua/opencode/cli/client.lua`:
- **In `handle_sse`**: Added check `#full_event > 0` before decoding; errors now use `DEBUG` level and respect `vim.g.opencode_debug`
- **In `handle_json`**: Added check `#full_data > 0` before decoding; errors now use `DEBUG` level and respect `vim.g.opencode_debug`

## Testing
Test by rapidly pressing `Ctrl-U`/`Ctrl-D` in opencode TUI. Errors should no longer appear unless `vim.g.opencode_debug` is enabled.

### DIFF
--- a/lua/opencode/cli/client.lua
+++ b/lua/opencode/cli/client.lua
@@ -4,11 +4,8 @@
 local M = {}
 
 local sse_state = {
-  -- Important to track the port, not just true/false,
-  -- because opencode may have restarted (usually on a new port) while the plugin is running.
+  -- Track the port - `opencode` may have restarted, usually on a new port
   port = nil,
-  -- Persistent buffer for SSE lines, because SSEs can span multiple `on_stdout` calls.
-  buffer = {},
   job_id = nil,
 }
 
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Good catch! Hmm, I think a race condition is the root cause here — `curl` calls share buffer state and it seems that the response from one call can get accidentally appended + processed as part of another call.

Scoping the buffer to each `curl` call fixed the `handle_json` path for me, but not `handle_sse`. I'm probably misunderstanding some pass by reference vs value here. I'll keep investigating, or let me know if you have any luck!

nickjvandyke:
I made some tweaks to your PR accordingly! I chose to leave the error notification in because I think off-by-default will result in too many silent failures.

Thanks again! 😀

Issue #104, opened 2025-12-26
feature: sync `opencode`'s CWD with Neovim's


horaoen:
### Did you check docs and existing issues?

- [x] I have read all the plugin docs
- [x] I have searched the existing issues
- [x] I have searched the existing [opencode issues](https://github.com/sst/opencode/issues)

### Operating system/version

Macos

### `:checkhealth opencode` output

opencode.nvim ~ - ✅ OK `nvim` version: `0.11.5`. - ✅ OK `opencode.nvim` git commit hash: `dfca5bb214d78a600781d50da350238b3e6e2621`. - ✅ OK `vim.g.opencode_opts`: {}  opencode.nvim [binaries] ~ - ✅ OK `opencode` available with version `1.0.202`. - ✅ OK `curl` available. - ✅ OK `pgrep` available: it will be used to auto-find `opencode` if `vim.g.opencode_opts.port` is not set. - ✅ OK `lsof` available: it will be used to auto-find `opencode` if `vim.g.opencode_opts.port` is not set.  opencode.nvim [snacks] ~ - ✅ OK `snacks.input` is enabled: `ask()` will be enhanced. - ✅ OK `blink.cmp` is available: `opts.ask.blink_cmp_sources` will be registered in `ask()`. - ✅ OK `snacks.picker` is enabled: `select()` will be enhanced.  opencode.nvim [providers] ~ - ✅ OK Configured `opencode` provider: `snacks`. - ✅ OK The `snacks` provider is available. - ⚠️ WARNING The `kitty` provider is not available — KITTY_LISTEN_ON environment variable is not set.   - ADVICE:     - Enable remote control in `kitty`. - ⚠️ WARNING The `wezterm` provider is not available — `wezterm` executable not found in `$PATH`.   - ADVICE:     - Install `wezterm` and ensure it's in your `$PATH`. - ⚠️ WARNING The `tmux` provider is not available — Not running inside a `tmux` session.   - ADVICE:     - Launch Neovim inside a `tmux` session. - ✅ OK The `terminal` provider is available.

### Describe the bug

When opencode is launched, it uses the working directory from the moment Neovim starts (where the `nvim` command was executed), rather than tracking Neovim's actual current working directory. This causes a mismatch when:

1. Using `nvim <folder>` to open a different directory
2. Navigating to a project via a dashboard plugin (like alpha-nvim, dashboard-nvim, etc.)
[...]

PR #105, opened 2025-12-26
feat: implement project root detection for terminal providers

horanf:
fix bug #104 
## Summary
Fix opencode working directory mismatch when Neovim's actual project directory differs from where nvim was launched.

## Problem
When using nvim <folder> or navigating to a project via a dashboard plugin, opencode would start in the wrong directory because all providers were inheriting the shell's working directory rather than detecting the actual project root.

## Scenarios affected:
- nvim ~/projects/my-app from ~/ → opencode opened in ~/ instead of ~/projects/my-app
- Using dashboard plugins (alpha-nvim, dashboard-nvim) to open projects
- Changing directories with :cd after Neovim starts

## Solution
Added get_project_root() function with smart detection priority:

1. Neovim directory argument (nvim <folder>)
2. Git root (for git repositories)
3. Current buffer directory (fallback for non-git projects)
4. LSP workspace root (if available)
5. vim.fn.getcwd() (final fallback)

All 5 providers now use this function:
- terminal.lua - Added cwd option to jobstart()
- snacks.lua - Pass cwd to snacks.terminal options
- tmux.lua - Added -c flag to split-window
- wezterm.lua - Added --cwd flag to split-pane
- kitty.lua - Changed from --cwd=current to explicit path

### DIFF
--- a/lua/opencode/provider/init.lua
+++ b/lua/opencode/provider/init.lua
@@ -58,7 +58,8 @@ local M = {}
 function M.get_project_root()
   local cwd = vim.fn.getcwd()
 
-  local git_root = vim.fn.systemlist("git -C " .. vim.fn.shellescape(cwd) .. " rev-parse --show-toplevel 2>/dev/null")[1]
+  local git_cmd = "git -C " .. vim.fn.shellescape(cwd) .. " rev-parse --show-toplevel 2>/dev/null"
+  local git_root = vim.fn.systemlist(git_cmd)[1]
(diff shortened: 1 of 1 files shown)


horanf:
I have tested it in my environment, and scenarios 1, 2, and 3 all work so far. The implementation for scenario 3 is to close the existing OpenCode workspace and create a new one.
### Scene 3
图片
图片
图片
then toggle again
图片

### scene 1、2
图片
图片
图片
图片
图片

PR #107, opened 2025-12-28
fix: prevent extmark end_col from exceeding line length

syscode1:
## Summary

Fix `nvim_buf_set_extmark` error when highlighting visual selections that extend to or beyond the end of a line.

## Problem

The `highlight()` function in `context.lua` calculates `end_col` as `range.to[2] + 1`, but this can exceed the actual line length, causing Neovim to throw an error:

```
E5108: Error executing lua ...opencode.nvim/lua/opencode/context.lua:75: in function 'nvim_buf_set_extmark'
```

This occurs when:
- The cursor/selection is at the end of a line
- The selection ends on an empty line
- Virtual column positions extend beyond line content

## Solution

Clamp `end_col` to the actual line length using `math.min(range.to[2] + 1, #line)`.

## Testing

Tested by selecting code and pressing `go` (operator mode) - the error no longer occurs.

### DIFF
--- a/lua/opencode/context.lua
+++ b/lua/opencode/context.lua
@@ -72,9 +72,15 @@ end
 ---@param buf integer
 ---@param range opencode.context.Range
 local function highlight(buf, range)
+  local end_row = range.to[1] - (range.kind == "line" and 0 or 1)
+  local end_col = nil
+  if range.kind ~= "line" then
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Ah thank you!! Honestly I mostly guess-and-checked when first adding this 😂 So I appreciate you catching this edge case and submitting a fix 🙂

Issue #97, opened 2025-12-18
bug: neovim visual selection highlight is not cleared when cancelling "ask"


lungsin:
### Did you check docs and existing issues?

- [x] I have read all the plugin docs
- [x] I have searched the existing issues
- [x] I have searched the existing [opencode issues](https://github.com/sst/opencode/issues)

### Operating system/version

CachyOS/rolling. Also happens in MacOS Sequoia 15.6.1

### `:checkhealth opencode` output

 ============================================================================== opencode:                                                                 3 ⚠️  opencode.nvim ~ - ✅ OK `nvim` version: `0.11.5+v0.11.5`. - ✅ OK `opencode.nvim` git commit hash: `39a246b597d6050ca319142b5af5a8b81c74e7d9`. - ✅ OK `vim.g.opencode_opts`: {}  opencode.nvim [binaries] ~ - ✅ OK `opencode` available with version `1.0.152`. - ✅ OK `curl` available. - ✅ OK `pgrep` available: it will be used to auto-find `opencode` if `vim.g.opencode_opts.port` is not set. - ✅ OK `lsof` available: it will be used to auto-find `opencode` if `vim.g.opencode_opts.port` is not set.  opencode.nvim [snacks] ~ - ✅ OK `snacks.input` is enabled: `ask()` will be enhanced. - ✅ OK `blink.cmp` is available: `opts.ask.blink_cmp_sources` will be registered in `ask()`. - ✅ OK `snacks.picker` is enabled: `select()` will be enhanced.  opencode.nvim [providers] ~ - ✅ OK Configured `opencode` provider: `snacks`. - ✅ OK The `snacks` provider is available. - ⚠️ WARNING The `kitty` provider is not available — KITTY_LISTEN_ON environment variable is not set.   - ADVICE:     - Enable remote control in `kitty`. - ⚠️ WARNING The `wezterm` provider is not available — Not running inside a `wezterm` window.   - ADVICE:     - Launch Neovim inside a `wezterm` window. - ⚠️ WARNING The `tmux` provider is not available — Not running inside a `tmux` session.   - ADVICE:     - Launch Neovim inside a `tmux` session. - ✅ OK The `terminal` provider is available.

### Describe the bug

After cancelling the UI prompt from `require('opencode').ask()` with some visual selection, the highlight from the visual selection is not cleared. Check the screenshots below for further details.

- When triggering `ask` command with a visual selection:
[...]

PR #110, opened 2026-01-03
fix: only restore visual mode selection if input was spawned from visual mode

s0ber:
The fix for this issue https://github.com/NickvanDyke/opencode.nvim/issues/97 was actually not enough. We should only restore visual selection if we were in a visual mode in a first place.

The issue is described in more details in this comment:
https://github.com/NickvanDyke/opencode.nvim/issues/97#issuecomment-3703980144

**PS.** Huge thanks for the great plugin.

### DIFF
--- a/lua/opencode/context.lua
+++ b/lua/opencode/context.lua
@@ -104,7 +104,7 @@ end
 
 function Context:resume()
   self:clear()
-  if self.range then
+  if self.range ~= nil then
     vim.cmd("normal! gv")
   end
 end


nickjvandyke:
oops thank you! not sure how I missed this 🤦 

Does it work to just check `self.range ~= nil`, rather than store this new field? Or does that miss an edge case?

s0ber:
@NickvanDyke It seems like it really should be enough! I've pushed a fix for that.

nickjvandyke:
yay, tyvm for this!

PR #113, opened 2026-01-05
Feature/custom chat frontend

chhlga:


nickjvandyke:
Thanks for this neat idea! However per the end of the README, I made a conscious decision to leverage `opencode`'s TUI:

> Uses opencode's TUI for simplicity — see [sudo-tee/opencode.nvim](https://github.com/sudo-tee/opencode.nvim) for a Neovim frontend.

May I ask what this adds over the plugin linked above? It's hard to tell from just the code.

PR #115, opened 2026-01-06
fix: don't expect buffer to be loaded to add it to context

s0ber:
I am using a quickfix list in a following way:

- I use nvim-tree plugin, where I can mark multiple files with `m` key
- I have a special command which sends all of the marked files in a quickfix list (I use Ctrl+q hotkey for that, similarly to Telescope plugin) 
- Then I can make those ones available to opencode with `@quickfix` context alias

The issue: currently `opencode.nvim` expects those buffers to be loaded. Which is not guaranteed with such workflow. Here I try to retain this check in other scenarios, but I've removed it from the one which is used by a quickfix.

I hope I've explained it clearly!

### DIFF
--- a/lua/opencode/context.lua
+++ b/lua/opencode/context.lua
@@ -24,7 +24,7 @@ local function last_used_valid_win()
   local latest_last_used = 0
   for _, win in ipairs(vim.api.nvim_list_wins()) do
     local buf = vim.api.nvim_win_get_buf(win)
-    if vim.api.nvim_buf_is_loaded(buf) and is_buf_valid(buf) then
+    if is_buf_valid(buf) then
       local last_used = vim.fn.getbufinfo(buf)[1].lastused or 0
       if last_used > latest_last_used then
         latest_last_used = last_used
@@ -309,7 +309,7 @@ function Context:visible_text()
   local visible = {}
   for _, win in ipairs(vim.api.nvim_list_wins()) do
     local buf = vim.api.nvim_win_get_buf(win)
-    if vim.api.nvim_buf_is_loaded(buf) and is_buf_valid(buf) then
+    if is_buf_valid(buf) then
       local start_line = vim.fn.line("w0", win)
       local end_line = vim.fn.line("w$", win)
       table.insert(


nickjvandyke:
ohh, neat! Thanks for fixing this.

Do you think we could/should remove the is_loaded check entirely? I wrote that a while ago and very well could have been wrong.

s0ber:
@NickvanDyke I'm not a big opencode specialist yet, so I'm not exactly sure when opencode.nvim should really expect file to be opened in NeoVim. E.g. I'm not yet sure in which cases opencode asks permission to edit a file, and when it just edits without asking anything. Also, if we expect opencode.nvim to run some commands in opened buffers, then maybe it should check if they are loaded right before running those commands. But, in most cases this happens BEFORE sending commands to opencode. So, once opencode will try to do some changes to files, their respective buffers could have already been closed.

So, even though there is this "is_loaded" check, it's not very bulletproof, and can easily break in some cases. So probably it can be removed.

s0ber:
@NickvanDyke ALSO, AFAIK quickfix entries have their own kind of buffer objects for every item in a list. We can get some info from them, but at the same time, they don't have to be loaded/opened.

nickjvandyke:
I think the concept of "loaded" is only relevant in Neovim - `opencode` doesn't care about it.

`is_buf_valid` primarily ensures that we only send actual content to `opencode`. For instance, without it, `@this` would refer to the buffer that we're currently `ask()`ing a prompt in. Not the most recent content-ful buffer.

So from that perspective, `is_loaded` seems irrelevant?

nickjvandyke:
So if that makes sense, let's change your PR to just removing the `is_loaded` check from `is_buf_valid` - without adding it separately elsewhere in the file - and see how that goes!

s0ber:
@NickvanDyke So.. Two remaining places where `nvim_buf_is_loaded` is checked were actually for window objects. And it turned out that if you have a window, it implies that its' buffer is loaded. So, this check in those places is completely unnecessary.

I've tested things and everything seem to work fine. But I would be glad for some additional testing, as I'm clearly not that familiar with the plugin.

**upd**: the case you described with the "last valid buffer" seems to work fine! E.g. I can spawn `ask` dialog from `nvim-tree` buffer, but still it would ask in a context of a last proper file-related buffer.

nickjvandyke:
Tyvm!

PR #119, opened 2026-01-07
fix(tmux): use pane PID to find correct opencode server port

### LINKED ISSUE #118: bug: OpenCode.nvim doesn't see opencode process started by itself (by this plugin)

### Operating system/version

MacOS 26.2

1 ==============================================================================    2 opencode:                                                                 2 ⚠️    3    4 opencode.nvim    5 - ✅ OK nvim version: 0.11.5.    6 - ✅ OK opencode.nvim git commit hash: 93567824fb171d0740d0f7f1e85118bb061f1811.    7 - ✅ OK vim.g.opencode_opts: {    8     provider = {    9       enabled = "tmux",   10       tmux = {   11         options = "-h -p 35"   12       }   13     }   14   }   15   16 opencode.nvim [binaries]   17 - ✅ OK opencode available with version 1.0.220.   18 - ✅ OK curl available.   19 - ✅ OK pgrep available.   20 - ✅ OK lsof available.   21   22 opencode.nvim [snacks]   23 - ✅ OK snacks.input is enabled: ask() will be enhanced.   24 - ✅ OK snacks.picker is enabled: select() will be enhanced.   25   26 opencode.nvim [providers]   27 - ✅ OK Configured opencode provider: tmux.   28 - ✅ OK The snacks provider is available.   29 - ⚠️ WARNING The kitty provider is not available — KITTY_LISTEN_ON environment variable is not set.   30   - ADVICE:   31     - Enable remote control in kitty.   32 - ⚠️ WARNING The wezterm provider is not available — wezterm executable not found in $PATH.   33   - ADVICE:   34     - Install wezterm and ensure it's in your $PATH.   35 - ✅ OK The tmux provider is available.   36 - ✅ OK The terminal provider is available.

### Describe the bug

I have multiple tmux sessions with nvim and opencode.nvim. And for some instances - all it fine, but for one - it just can't find openned opencode by themself. Restart doesn't work. All commands sent to opencode from nvim just doesn't have any effect.

### Steps To Reproduce

1. Hard to say: it doesn't work at one place. And I don't know how to troubleshoot.

### Expected Behavior

Should be connected always - especially if opencode was started by the plugin.

PhantomYdn:
When multiple opencode instances share the same CWD, the previous CWD-based discovery could select the wrong instance. This was particularly problematic with tmux provider since opencode processes spawned via tmux are not descendants of Neovim.

Changes:
- Add get_port() method to provider interface
- Implement get_port() for tmux provider using pane PID tracing
- Add process utilities (get_descendants, get_listening_port)
- Check provider's get_port() before falling back to CWD discovery
- Other providers return nil (fallback to existing behavior)

Fixes #118

### DIFF
--- a/lua/opencode/cli/server.lua
+++ b/lua/opencode/cli/server.lua
@@ -297,6 +297,19 @@ function M.get_port(launch)
   local Promise = require("opencode.promise")
 
   return Promise.new(function(resolve, reject)
+    -- Check if provider can supply port directly
+    local provider = require("opencode.config").provider
+    if provider and provider.get_port then
+      local provider_port = provider:get_port()
+      if provider_port then
+        local ok, _ = pcall(test_port, provider_port)
+        if ok then
+          resolve(provider_port)
+          return
+        end
+      end
+    end
+
     local configured_port = require("opencode.config").opts.port
     local find_port_fn = configured_port and function()
       return test_port(configured_port)
--- a/lua/opencode/provider/init.lua
+++ b/lua/opencode/provider/init.lua
@@ -31,6 +31,10 @@
 ---Should return `true` if the provider is available,
 ---else an error string and optional advice (for `vim.health.warn`).
 ---@field health? fun(): boolean|string, ...string|string[]
+---
+---Get the port of the `opencode` server started by this provider.
+---Returns `nil` if the provider cannot determine the port (falls back to CWD-based discovery).
+---@field get_port? fun(self: opencode.Provider): number|nil
 
 ---Configure and enable built-in providers.
 ---@class opencode.provider.Opts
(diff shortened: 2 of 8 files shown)


nickjvandyke:
Thanks for proofing this out! I think it's the correct solution for #118. I hope to review it soon.

I'm also curious which other providers could surface their `opencode` PID. Not necessarily for this PR, but I'd like this pattern to be generally useful if we incorporate it.

nickjvandyke:
Closing because the approach and code surrounding this has changed so much. We do still have #164. But this PR is for tmux, which we no longer officially support for maintenance reasons.

Sorry! But thank you for the discussion and effort 🙏

Issue #106, opened 2025-12-28
feature: OpenCode formatters path when using mason


jcromero:
### Did you check the docs?

- [x] I have read all the docs

### Is your feature request related to a problem? Please describe.

How are you coping with the OpenCode formatters path when using mason in neovim? It is not detected by an embedded OpenCode launched by this plugin.

### Describe the solution you'd like

Currently the only solution I can think of is to configure the `command` property one by one in my project opencode.json file or am I missing anything?

### Describe alternatives you've considered

I have no alternative.

nickjvandyke:
Oh interesting, I didn't know `opencode` does this!

As I understand `mason`, at runtime it adds its installed stuff to `$PATH`. So maybe we do the same with `opencode`? e.g. with `snacks.terminal`, something like:

```lua
vim.g.opencode_opts = {
  provider = {
    snacks = {
      env = {
        PATH = vim.env.PATH .. ":" .. require("mason").get_install_path() -- I presume mason exposes something like this
      }
    }
  }
}
```

nickjvandyke:
I think this works already? I have `stylua` installed only via Mason, and an `opencode` embedded in both the "snacks" and "terminal" providers successfully runs it when I ask.

Gonna close because I haven't heard anything in a while, but let me know if this turns out incorrect!

jcromero:
I think you are running the built-in formatter embedded in opencode, not the mason one

nickjvandyke:
I don't think so - when I run `opencode` outside Neovim and ask it to run `stylua`, I get a "command not found" error:

<img width="836" height="1015" alt="Image" src="https://github.com/user-attachments/assets/7960d3a9-02d4-4e40-92c0-828e2aaab1bb" />

Unless there is a difference between asking `opencode` to run the command, and it automatically calling the formatter itself?

nickjvandyke:
Here is the successful `stylua` call when running `opencode` inside Neovim:

<img width="1566" height="1015" alt="Image" src="https://github.com/user-attachments/assets/c5d1318e-4749-4402-8209-d16537704aa0" />

jcromero:
Interesting, I don't know how does it find the command, because if I try to exec "! stylua" from inside opencode it does not find anything

nickjvandyke:
Hmm, `! stylua` works for me in both the `snacks` and `terminal` providers. Are you sure `mason` is set up before you open the embedded `opencode`? So that it's added all its stuff to `$PATH`.

[...]

PR #126, opened 2026-01-13
Fix opencode command for kitty provider

phantomwhale:
Kitty requires the executable and any extra arguments to be provided as an array of values. Before this change it would try and execute "opencode --port" and state no such executable could be found

This fixes the issue for the kitty provider; unsure if other providers would need a similar fix

### DIFF
--- a/lua/opencode/provider/kitty.lua
+++ b/lua/opencode/provider/kitty.lua
@@ -131,7 +131,10 @@ function Kitty:start()
     table.insert(launch_cmd, "--location=" .. location)
   end
 
-  table.insert(launch_cmd, self.cmd)
+  -- Split cmd string into separate arguments for kitty launch
+  for arg in self.cmd:gmatch("%S+") do
+    table.insert(launch_cmd, arg)
+  end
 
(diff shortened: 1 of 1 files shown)


phantomwhale:
Whilst I'm here, I've also used this change to let me add `--copy-env` to my provider config (e.g. cmd = "--copy env opencode --port") which means kitty opens up the opencode subshell with the current environment; without that, opencode can't run my test suites etc... as it doesn't have the right executables on it's PATH

Not sure if that would be something that _should_ be part of this plugin by default, or if it's something deliberately left as an opt-it thing. Without this PR, I've simply not been able to do it at all (aside from hacking the option into my local copy of the kitty provider) hence I thought to raise it here as an FYI.

nickjvandyke:
> Whilst I'm here, I've also used this change to let me add --copy-env to my provider config (e.g. cmd = "--copy env opencode --port") which means kitty opens up the opencode subshell with the current environment; without that, opencode can't run my test suites etc... as it doesn't have the right executables on it's PATH

Oh neat, maybe that's the correct fix for #106 ?

nickjvandyke:
Ty!

PR #127, opened 2026-01-13
refactor(ui): simplify select rendering to avoid alignment issues

epheien:
## Summary
- Simplified the select rendering logic in `lua/opencode/ui/select.lua`
- Removed unnecessary indentation calculations that were causing alignment issues
- The change makes the code cleaner and more maintainable

## Changes
- Removed the `indent` variable calculation based on item index
- Simplified group divider rendering to remove unnecessary spacing
- Simplified regular item formatting to remove the leading spacing

This refactoring improves the visual alignment of the select UI and reduces code complexity.

For example, when using dressing+telescope as the implementation of vim.ui.select, this situation occurs.

Xnip2026-01-13_03-00-44

### DIFF
--- a/lua/opencode/ui/select.lua
+++ b/lua/opencode/ui/select.lua
@@ -183,17 +183,11 @@ function M.select(opts)
             table.insert(formatted, 2, { string.rep(" ", 18 - #item.name) })
             return formatted
           else
-            local indent = #tostring(#items) - #tostring(item.idx)
             if item.__group then
               local divider = string.rep("—", (80 - #item.name) / 2)
-              return string.rep(" ", indent) .. divider .. item.name .. divider
+              return divider .. item.name .. divider
             end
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Ah, well this indentation is for the default `vim.ui.select` because it prepends the index number to each row. But I forgot about other `vim.ui.select` replacements. I wonder how we could only do this for the former case 🤔

epheien:
> Ah, well this indentation is for the default `vim.ui.select` because it prepends the index number to each row. But I forgot about other `vim.ui.select` replacements. I wonder how we could only do this for the former case 🤔

I know the purpose of doing it that way. What I mean is, if it's using the native vim.ui.select, then there's no need for padding, because it's hard to tell whether vim.ui.select is using the native one or another substitute.

epheien:
I choose to modify the dressing code to solve this alignment issue.

PR #128, opened 2026-01-13
fix(cli): filter opencode processes by --port argument

epheien:
## Summary
- Fixed the process filtering logic in `lua/opencode/cli/server.lua`
- Now correctly filters opencode processes by checking for the `--port` argument
- This prevents false positives when multiple processes contain "opencode" in their command line

## Changes
- Updated the `pgrep` command to include `-lfa` flags and filter by `--port` argument
- Changed PID extraction to properly parse the new `pgrep -l` output format

This fix ensures that only actual opencode server processes (those started with `--port` argument) are discovered, improving reliability when multiple processes with similar names are running.

### DIFF
--- a/lua/opencode/cli/server.lua
+++ b/lua/opencode/cli/server.lua
@@ -37,7 +37,10 @@ local function get_processes_unix()
   assert(vim.fn.executable("pgrep") == 1, "`pgrep` executable not found")
   assert(vim.fn.executable("lsof") == 1, "`lsof` executable not found")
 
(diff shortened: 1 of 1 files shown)


epheien:
The motivation for this modification is that under macOS, there are many opencode processes, which causes this plugin to be very slow when acquiring ports.

Xnip2026-01-13_23-42-55

nickjvandyke:
Thanks for this! I guess the way I use `opencode` doesn't result in nearly as many processes (last I checked), but this seems great for some users! 😀

PR #134, opened 2026-01-16
feat(context): add `absolute_paths` config option

Caylub:
## Summary
Add `absolute_paths` config option for file path formatting in contexts.

## Problem
The plugin allows connecting to opencode instances in subdirectories of 
Neovim's CWD, but relative paths don't resolve correctly in that case.
When opencode wasn't in the exact same directory as Neovim, the model 
returned gibberish because it couldn't read the referenced files.

## Solution
New `absolute_paths = true` option uses full paths instead of CWD-relative paths.

## Usage
```lua
vim.g.opencode_opts = { absolute_paths = true }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,6 +91,18 @@ programs.nixvim = {
 | `@marks`         | Global marks                                                |
 | `@grapple`       | [grapple.nvim](https://github.com/cbochs/grapple.nvim) tags |
 
+#### Absolute Paths
+
+By default, file paths in contexts are relative to Neovim's working directory. If `opencode` is running in a different directory, these relative paths may not resolve correctly.
+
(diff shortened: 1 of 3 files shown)


nickjvandyke:
Thanks for surfacing this!

I wonder if this depends on the model? IME `opencode` succesfully ignores the "extra" dirs at the beginning of the path.

IMO we should solve this without additional, conditional configuration. We could either:

1. Only find `opencode`s that exactly match Neovim's CWD. I only allowed sub-dirs because it initially seemed to work for me.
2. Send file paths relative to `opencode`'s CWD (which we now have), instead of Neovim's.

I like the simplicity and predictability of 1, but idk if that'd break anyone's workflow. It'd also work better after #104

Issue #101, opened 2025-12-20
Feature Request: Partial Accept and Diff View


nesharp:
### Did you check the docs?

- [x] I have read all the docs

### Is your feature request related to a problem? Please describe.

Would love to use this for real development work, but currently missing two essential features that Cursor has:

1. **Partial Accept** - ability to accept code suggestions line-by-line or in chunks
2. **Diff View** - visual comparison showing what's changing (side-by-side or inline)

## Why This Matters

Right now, I have to accept entire suggestions without being able to:
- Review changes visually before applying
- Accept only the good parts and skip others
- See exactly what lines are being modified

This makes it risky for production code where I need to carefully control what gets changed.

## What Would Help

- Side-by-side or inline diff display (like `git diff`)
- Ability to navigate through changes and accept/reject individually
- Clear visual indicators of additions/deletions/modifications

## Personal Note

I'm really rooting for this project! This is literally the **only reason** I'm not using nvim full-time right now. I'd switch immediately if these features were available. Would absolutely love to see this implemented - the potential is huge! 🚀

### Describe the solution you'd like

-

### Describe alternatives you've considered

-

nickjvandyke:
I agree it'd be neat! I haven't done this yet because it's a lot of work and somewhat duplicates `opencode`'s permissions.

Have you seen https://github.com/NickvanDyke/opencode.nvim/pull/91 ?

nickjvandyke:
I've pushed an experimental implementation to `main` - please try it out and let me know what you think! I surely missed some edge cases. See https://github.com/nickjvandyke/opencode.nvim#edits-1 for details.

`opencode` does not natively support partial edit permission accepts/rejects (afaik). But you can use Neovim's native `dp` and `do` keymaps to natively accept/reject individual hunks. To prevent conflicts, the plugin then rejects the edit request. I think that's the best we can do for partials, but I'm open to better workarounds.

PR #137, opened 2026-01-19
feat: add session-wise diff support using codediff.nvim

naowalrahman:
Using [codediff.nvim](https://github.com/esmuellert/codediff.nvim), this PR enables users to pick a session to view a diff of. It compares the state of the repository at the start of the session to that after the session by somewhat reverse-engineering using opencode's snapshots feature. Since that feature only works on git repositories, this diff feature also only works on git repositories. 

### Changes
- adds `session diff` to `require("opencode").select` menu, which can be toggled via select opts 
- warns users if trying to run `require("opencode").session_diff` without codediff.nvim installed
- uses `vim.ui.select` to let user pick which session to diff, fetched from `/session` endpoint of opencode server
- creates before/after temp directories for the file states from the snapshots and diffs them with codediff.nvim 

Addresses #101 and #91. 

### Screenshots

screenshot_2026-01-18_09-26-24-PM
screenshot_2026-01-18_09-28-09-PM

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -158,6 +158,10 @@ local defaults = {
     },
     tmux = {
       options = "-h", -- Open in a horizontal split
+      focus = false, -- Keep focus in Neovim
+      -- Disables allow-passthrough in the tmux split
+      -- preventing OSC escape sequences from leaking into the nvim buffer
+      allow_passthrough = false,
     },
   },
 }
--- a/lua/opencode/provider/tmux.lua
+++ b/lua/opencode/provider/tmux.lua
@@ -13,6 +13,25 @@ Tmux.name = "tmux"
 ---
 ---`tmux` options for creating the pane.
 ---@field options? string
+---
+---Focus the opencode pane when created. Default: `false`
+---@field focus? boolean
+--
+---Allow `allow-passthrough` on the opencode pane.
+-- When enabled, opencode.nvim will use your configured tmux `allow-passthrough` option on its pane.
+-- This allows opencode to use OSC escape sequences, but may leak escape codes to the buffer
+-- (e.g., "=31337;OK" appearing in your buffer).
+--
+-- Limitations of having allow-passthrough disabled in the opencode pane:
+-- - can't display images
+-- - can't use special (terminal specific; non-system) clipboards
+-- - may have issues setting window properties like the title from the pane
+--
+-- If you enable this, consider also enabling `focus` to auto-focus the pane on creation,
+-- which can help avoid OSC code leakage while opencode is sending escape sequences on startup.
+--
+-- Default: `false` (allow-passthrough is disabled to prevent OSC code leakage)
+---@field allow_passthrough? boolean
 
 ---@param opts? opencode.provider.tmux.Opts
 ---@return opencode.provider.Tmux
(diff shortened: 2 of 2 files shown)


naowalrahman:
P.S. there's a type checking error between `string|nil` and `string` in `diff.lua` that I tried to mess around with but couldn't quite fix, so I'd appreciate any help fixing that 😅

naowalrahman:
A few changes with the last 3 commits: 
- fixed the LSP error so all checks pass now 
- removed temp directory cleanup all together because they get erased when you close neovim anyway
- allow users to set their opencode data directory in opts, eliminating cross-platform storage directory differences

naowalrahman:
I just realized that opencode exposes a /session/:id/diff [endpoint](https://opencode.ai/docs/server/#:~:text=GET-,/session/%3Aid/diff,-Get%20the%20diff) to get an array of FileDiff objects, which simplifies things much more and doesn't require reverse-engineering snapshots. 

@NickvanDyke would you be interested in a feature like this? Happy to work on it if so.

nickjvandyke:
@naowalrahman I'm hesitant to accept this because
1. I don't see the use case for viewing the session's internal diff? And specifically in Neovim. But that could just be my workflow.
2. It seems dependent on `opencode`'s internals, i.e. prone to breaking changes
3. Using codediff.nvim is a bit specific to include in core functionality.

_If_ there's a strong community desire for 1, then I could be open to a more general solution (addressing 2 and 3).

(But I appreciate your effort regardless! 🙂 )

naowalrahman:
> 1. I don't see the use case for viewing the session's internal diff? And specifically in Neovim. But that could just be my workflow.

@NickvanDyke I'm a bit fond of the inline diff functionality from Cursor/Zed where each set of updates from the agent accumulates as inline diffs that can be accepted/rejected per-hunk. It's probably because I dont commit very often or sometimes I'm vibe coding a proof of concept where I want to get something working before committing. That's mainly why I wanted some sort of session-level or even message-level diff. 

But yes - it depends on the community desire. The other opencode.nvim has a session-level diff feature using the native `vim.diff` that (I think) uses the `/session/:id/diff` endpoint, so that's a good place to look.

Thank you for the great plugin nonetheless!

nickjvandyke:
I intend for this plugin more to supplement your editor workflow. So things like per-message diffs (like #133) align better with that.

I also would not be surprised if `opencode` eventually adds session-level diff viewing, or better integrates with some external tool like `lazygit`. So I'm trying not to duplicate that work here.

> I'm a bit fond of the inline diff functionality from Cursor/Zed where each set of updates from the agent accumulates as inline diffs that can be accepted/rejected per-hunk

I don't think this is possible with `opencode` right? As cool as it sounds. Because it blocks, waiting for us to accept/reject the proposed diff. And the workaround to accumulate them ourselves is too close to what git tools already do imo.

naowalrahman:
> I intend for this plugin more to supplement your editor workflow. So things like per-message diffs (like #133) align better with that.
> I also would not be surprised if `opencode` eventually adds session-level diff viewing, or better integrates with some external tool like `lazygit`. So I'm trying not to duplicate that work here.

Sounds good! You're probably right; there's an open issue (anomalyco/opencode#5102). 

> > I'm a bit fond of the inline diff functionality from Cursor/Zed where each set of updates from the agent accumulates as inline diffs that can be accepted/rejected per-hunk
> 
> I don't think this is possible with `opencode` right? As cool as it sounds. Because it blocks, waiting for us to accept/reject the proposed diff. And the workaround to accumulate them ourselves is too close to what git tools already do imo.

The waiting for accept/reject is user-dependent because you can set set edit tools to not request permission in opencode.json. But yeah, it is quite a workaround. I will wait for #133!

nickjvandyke:
> you can set set edit tools to not request permission in opencode.json

Right, I meant when listening for `opencode`'s `edit` permission SSE 🙂 Without which we have to diff stuff ourselves rather than leverage existing stuff.

PR #138, opened 2026-01-19
feat(fzf): add fzf-lua integration for buffer and file selection

xeon826:
File picker using fzf-lua to select git files or open buffers to add to the prompt. Use it like so:

```
		vim.keymap.set("t", "<C-f>", function()
			require("opencode").fzf.select_files()
		end, opts)
		vim.keymap.set("t", "<C-b>", function()
			require("opencode").fzf.select_buffers()
		end, opts)
```
image

### DIFF
--- a/lua/opencode/ui/fzf.lua
+++ b/lua/opencode/ui/fzf.lua
@@ -6,6 +6,14 @@ local function is_buf_valid(buf)
   return vim.api.nvim_get_option_value("buftype", { buf = buf }) == "" and vim.api.nvim_buf_get_name(buf) ~= ""
 end
 
+local function escape_terminal_mode()
+  local mode = vim.api.nvim_get_mode().mode
+  if mode:match("^[it]") then
+    vim.cmd("stopinsert")
+    vim.fn.feedkeys("\27", "n")
+  end
+end
+
 local function strip_ansi_codes(str)
   local result = str
   result = result:gsub("\27%[[0-9;]*m", "")
@@ -54,6 +62,8 @@ local function send_files_to_opencode(files, prompt_prefix)
 end
 
 function M.select_buffers(prompt_prefix)
+  escape_terminal_mode()
+
   local ok, fzf = pcall(require, "fzf-lua")
   if not ok then
     vim.notify("fzf-lua not found. Please install fzf-lua to use this feature.", vim.log.levels.ERROR)
@@ -109,6 +119,7 @@ function M.select_buffers(prompt_prefix)
 end
 
 function M.select_files(prompt_prefix, opts)
+  escape_terminal_mode()
   opts = opts or {}
 
   local ok, fzf = pcall(require, "fzf-lua")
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Neat, thank you! I have thought about something like this.

I think it'd be more powerful and reusable [the way sidekick.nvim does it](https://github.com/folke/sidekick.nvim#snacksnvim-picker-integration). Is that possible in fzf-lua? That could get us started on supporting other pickers too. 🙂

xeon826:
I created [another branch using my fork](https://github.com/xeon826/opencode.nvim/tree/file-picker) that more resembles the functionality you're referring to but it seems opencode is lacking an endpoint for submitting file paths so [I created an issue and rough draft branch of opencode for adding that.](https://github.com/anomalyco/opencode/issues/9643)

nickjvandyke:
I think that missing feature is fine for now - it's an optimization, not a requirement. And is consistent with other filepath references anyway.

naowalrahman:
> Neat, thank you! I have thought about something like this.
> 
> I think it'd be more powerful and reusable [the way sidekick.nvim does it](https://github.com/folke/sidekick.nvim#snacksnvim-picker-integration). Is that possible in fzf-lua? That could get us started on supporting other pickers too. 🙂

Just came up with a neat snacks.picker() action to send context from snacks.picker to opencode, similar to the sidekick.nvim integration:
```lua
return {
  "folke/snacks.nvim",
  opts = {
    picker = {
      actions = {
        ---@param picker snacks.Picker
        opencode_send = function(picker)
          local context = {}

          for _, item in ipairs(picker:selected { fallback = true }) do
            local entry = ""

            -- Build file reference with optional line range
            if item.file then
              entry = "@" .. item.file
              if item.pos then
                local has_range = item.end_pos and item.end_pos[1] ~= item.pos[1]
                entry = entry .. (" L%d"):format(item.pos[1])
                if has_range then entry = entry .. ("-L%d"):format(item.end_pos[1]) end
              end
            end

            -- Append text content
            if item.text and item.text ~= "" then
              entry = entry .. (item.file and "\n  " or "") .. item.text
            end

            context[#context + 1] = entry
          end

          require("opencode").prompt(table.concat(context, "\n") .. "\n")
        end,
      },
      win = {
        input = {
          keys = {
            ["<a-a>"] = {
              "opencode_send",
              mode = { "n", "i" },
            },
          },
        },
      },
    },
  },
}
```

Press Alt+A on current item/selected items in snacks.picker to send context to opencode. 

Not sure if fzf-lua supports actions like snacks.picker, but I think picker integrations can be handled on the picker-side like this instead of plugin-side.

nickjvandyke:
Nice @naowalrahman! Could you PR that action function? Maybe to `lua/opencode/integrations/picker/snacks.lua` or similar. Then I can add a README bit like `sidekick.nvim` has.

nickjvandyke:
@xeon826 the snacks.picker integration was merged in #152 if you care to use it as a reference to adapt your fzf-lua integration 🙂

nickjvandyke:
Closing as stale, but please re-open a new PR with the suggested approach if you feel inspired 🙂

Issue #123, opened 2026-01-09
feature: cycle through sessions


nkabbara:
### Did you check the docs?

- [x] I have read all the docs

### Is your feature request related to a problem? Please describe.

Hello. I can currently list sessions using `require("opencode").command("session.list")`, but I don't have a way to cycle through them.

### Describe the solution you'd like

It would be great if one could get a UI in their current buffer to select a session.

### Describe alternatives you've considered

Currently, I'm having to switch windows, select the session I want, then switch back.

nickjvandyke:
I don't think `opencode` exposes a "select session" endpoint unfortunately, so this would depend on them first. Even then, I try to minimize re-implementing TUI features for simplicity, both in code and UX.

nkabbara:
Thanks for the plugin @NickvanDyke. It's super useful.

I'll try and implement a local workaround in the meantime.

nickjvandyke:
Let us know what you come up with!

nickjvandyke:
Ope it turns out `opencode` does expose a "select session" event! https://github.com/anomalyco/opencode/blob/dev/packages/opencode/src/cli/cmd/tui/event.ts

And then you could fetch all of them from [`/sessions`](https://opencode.ai/docs/server/#sessions), to display in `vim.ui.select`.

If you get something working, feel free to open a PR!

nkabbara:
I did a PoC and it works great. Would we want to make something like this as part of `opencode.ui.select` or keep it separate?

I'm leaning towards putting it in its own `vim.ui.select` because we will probably keep adding features to session related tasks like rename, delete, etc...

![Image](https://github.com/user-attachments/assets/f63d9751-0a66-4290-b2c9-de9fc3dc6a4d)

nickjvandyke:
Ooo nice!

That's a good question. I initially designed `opencode.ui.select` to allow mix-and-matching of sections. So the default behavior shows everything, for ease-of-use/setup. But users can configure the keymap to only show a single section, to get the same behavior as though the section were its own `vim.ui.select`. So, thoughts on adding it as a new section? It will probably need a better abstraction as it grows, which I can look into.

PR #139, opened 2026-01-19
feat(sessions): select session

nkabbara:
Re: #123. One way to check this out is to open the main select dialogue and pick the select session command.

### DIFF
--- a/lua/opencode/ui/select_session.lua
+++ b/lua/opencode/ui/select_session.lua
@@ -1,5 +1,15 @@
 local M = {}
 
+local function ellipsize(s, max_len)
+  if vim.fn.strdisplaywidth(s) <= max_len then
+    return s
+  end
+  local truncated = vim.fn.strcharpart(s, 0, max_len - 3)
+  truncated = truncated:gsub("%s+%S*$", "")
+
+  return truncated .. "..."
+end
+
 function M.select_session()
   require("opencode.cli.server")
     .get_port()
@@ -21,7 +31,7 @@ function M.select_session()
         format_item = function(item)
           local title_length = 60
           local updated = os.date("%b %d, %Y %H:%M:%S", item.time.updated / 1000)
-          local title = M.ellipsize(item.title, title_length)
+          local title = ellipsize(item.title, title_length)
           return ("%s%s%s"):format(title, string.rep(" ", title_length - #title), updated)
         end,
       }, function(choice)
@@ -33,16 +43,6 @@ function M.select_session()
     :catch(function(err)
       vim.notify(err, vim.log.levels.ERROR)
     end)
-end
-
-function M.ellipsize(s, max_len)
-  if vim.fn.strdisplaywidth(s) <= max_len then
-    return s
-  end
-  local truncated = vim.fn.strcharpart(s, 0, max_len - 3)
-  truncated = truncated:gsub("%s+%S*$", "")
-
-  return truncated .. "..."
 end
 
 return M


nkabbara:
Hey @NickvanDyke!

What do you think about this? Any suggestions?

nickjvandyke:
Sorry just been busy! LGTM, thanks for slotting it in so nicely to the surrounding code + patterns!

I'm generally trying to avoid re-implementing TUI features to keep the plugin simple and reliable for both me and users. I'll accept this because the implementation is good and simple, and hopefully `opencode` doesn't make many breaking changes to this. But if you plan to add more TUI session features, it may be worth discussing first 🙂

nkabbara:
> Sorry just been busy! LGTM, thanks for slotting it in so nicely to the surrounding code + patterns!
> 
> I'm generally trying to avoid re-implementing TUI features to keep the plugin simple and reliable for both me and users. I'll accept this because the implementation is good and simple, and hopefully `opencode` doesn't make many breaking changes to this. But if you plan to add more TUI session features, it may be worth discussing first 🙂

Thanks for the kind words, appreciate it.

Makes sense. Do you have a roadmap on where you see this going as far as how it will interact with opencode?  The tui keeps adding ui features which will make the plugin harder to use if we don't implement them somehow. For example it's now starting to ask clarifying questions 😅 .

Screenshot 2026-01-24 at 13 59 48

nickjvandyke:
The plugin's first priority is enabling functionality that's not possible in the TUI. Supplementing your editor workflow with AI, rather than replace it. Like injecting editor context. Neovim implementations of TUI features are nice, but it's impossible for me to keep up with `opencode`'s pace. As your example demonstrates. They're a paid team, and I'm a single unpaid maintainer 😅 

Maintenance aside, I also want to keep things simple for users. Sometimes less is more. Maybe this is just me, but I found it overwhelming to try so many new AI tools. They all have their own way to do things. With `opencode.nvim`, I wanted it to be like "use `opencode` how you already know, _plus_ these useful new things." Not confuse users with a second way to do something they already know how to do in `opencode`.

Obviously there are exceptions. Like `opencode.nvim` can call `opencode` commands. But the complexity there is relatively low for the convenience and integration it enables. That tradeoff between complexity and power is the main thing I consider for "duplicative" features. While I appreciate any and all effort, I can't accept every contribution or suggestion because it would both dilute the plugin's purpose and overwhelm me, harming the quality of the plugin's core features.

This is my experience in open source thus far, at least 🙂 I would love to create the best editor experience possible (in Neovim and others), but it's only realistic as a paid member of the `opencode` team or something.

nkabbara:
> This is my experience in open source thus far, at least 🙂 I would love to create the best editor experience possible (in Neovim and others), but it's only realistic as a paid member of the opencode team or something.

Yeah, totally makes sense. How's your current workflow when something like the above happens? Do you ctrl-w to the opencode split and make the selection then go back to editing?

nickjvandyke:
> How's your current workflow when something like the above happens? Do you ctrl-w to the opencode split and make the selection then go back to editing?

I don't think I'm on the `opencode` version with that feature yet haha, but yes, generally switching to the `opencode` window and using their UI is the default "solution." I hoped the permission dialog would be a good general-purpose solution to such blocking events, but it seems they have sidestepped or enhanced that model.

nkabbara:
> > How's your current workflow when something like the above happens? Do you ctrl-w to the opencode split and make the selection then go back to editing?
> 
> I don't think I'm on the `opencode` version with that feature yet haha, but yes, generally switching to the `opencode` window and using their UI is the default "solution." I hoped the permission dialog would be a good general-purpose solution to such blocking events, but it seems they have sidestepped or enhanced that model.

I'm also curious about the opencode devs' workflow since they use vim.

nickjvandyke:
Interesting, I didn't know that! While they've nicely supported the APIs this plugin needs, they haven't interacted with it much otherwise. So I guess they prefer to use `opencode` in isolation? I suppose that gives a more accurate dogfooding perspective too.

nickjvandyke:
Code looks and functions well! Tyvm for the discussion and iteration 🙂 ❤️

PR #141, opened 2026-01-21
Use vim.system instead of io.popen

benjosua:
Replace io.popen() with vim.system() for safer command execution and better error handling.


nickjvandyke:
Nice, tyvm for this!

Now that we have more information about the result, could you check for a non-zero codes and `error(<cmd> failed with <code>\nstderr:\n<stderr>)` in that case? Just to save us from some silent failure headaches in the future 😀

PR #142, opened 2026-01-23
fix(client): use jobstart for SSE streaming

### LINKED ISSUE #44: bug: missing `file.edited` events from Opencode

### Neovim version (nvim -v)

0.11.3

### Operating system/version

NixOS 25.05

### Opencode version (opencode -v)

0.13.5

### Describe the bug

Opencode.nvim occasionally misses `file.edited` events from Opencode (cli), breaking buffer reload on change.

e-cal:
Fixes #44, previous fix seems not to always work and the issue persists.

Replaces vim.system with vim.fn.jobstart. jobstart streams data as it comes in while vim.system seems to buffer more aggressively occasionally missing events.

### DIFF
--- a/lua/opencode/cli/client.lua
+++ b/lua/opencode/cli/client.lua
@@ -101,6 +101,7 @@ local function curl(url, method, body, callback)
   end
 
   local stderr_lines = {}
+  -- Would prefer `vim.system`, but it seems unable to handle rapid SSEs currently and occasionally drops them.
   return vim.fn.jobstart(command, {
     on_stdout = function(_, data)
       if not data then


nickjvandyke:
Darn, I was excited to migrate to the newer vim.system haha. I guess I didn't test extensively enough.

> occasionally missing events

Missing entirely? Or just delayed? The former makes me think maybe I just introduced a bug while migrating. `vim.fn.jobstart` passes `data[]`, whereas `vim.system` is `data`. So I may have misunderstood how to adjust for that.

e-cal:
Missing entirely such that the buffer reload autocommand doesn't trigger, since while its processing the batch it isn't listening. It's dependent on how fast opencode is sending the events, but since `file.edited` is much less common than the message updated ones it gets buried and much more likely to be one of the missed events.

From the docs I think `Job_control` is also simply the API designed for this use case more so than `vim.system`

[Job_control](https://neovim.io/doc/user/job_control.html)
> Job control is a way to perform multitasking in Nvim, so scripts can spawn and control multiple processes without blocking the current Nvim instance.
> ...
> Job stdio streams form a [channel](https://neovim.io/doc/user/channel.html#channel) which can send and receive raw bytes or [msgpack-rpc](https://neovim.io/doc/user/api.html#msgpack-rpc) messages.

[`vim.system`](https://neovim.io/doc/user/lua.html#vim.system()) on the other hand uses [`uv.spawn`](https://neovim.io/doc/user/luvref.html#uv.spawn()) under the hood:
> Initializes the process handle and starts the process. If the process is successfully spawned, this function will return the handle and pid of the child process.

While it's meant to replace job control eventually, it seems to me that right now it is polling the stdout/stderr of the subprocess at set intervals rather than streaming in the raw data. 

So far I can't seem to reliably get around this with `vim.system` while `vim.fn.jobstart` works consistently and reliably.

nickjvandyke:
Thanks for the details! And the clean fix. I guess the code is at least a bit cleaner now than it was pre-vim.system, so still a win 🙂

PR #144, opened 2026-01-25
fix(tmux): add config options to prevent OSC response leak into Neovim buffer

e-cal:
Add config options for creating a split in tmux:

```lua
provider = {
  tmux = {
    focus = false,              -- Focus the opencode pane when created (default: false)
    disable_passthrough = true, -- Disable allow-passthrough on the pane (default: true)
  },
}
```

### Details

When tmux has `allow-passthrough on` configured, launching opencode writes `=31337;OK` to the buffer (an iTerm2 OSC 1337 response). 

Caused by opencode TUI querying terminal capabilities on startup with nvim pane still in focus.

#143 also aims to address this issue via auto-focus, so opencode eats the OSC response as its meant to.

Some users may prefer keeping the launch in the background, others may require allow-passthrough in the opencode pane. So, I propose making your choice of workaround a config option.

I make `disable_passthrough = true` by default though since this is the less invasive workaround and most users won't notice the difference anyway.

### DIFF
--- a/lua/opencode/provider/tmux.lua
+++ b/lua/opencode/provider/tmux.lua
@@ -19,8 +19,8 @@ Tmux.name = "tmux"
 --
 ---Allow `allow-passthrough` on the opencode pane.
 -- When enabled, opencode.nvim will use your configured tmux `allow-passthrough` option on its pane.
--- This allows opencode to use OSC escape sequences, but may leak escape codes to the terminal
--- (e.g., "=31337;OK" appearing in your terminal output).
+-- This allows opencode to use OSC escape sequences, but may leak escape codes to the buffer
+-- (e.g., "=31337;OK" appearing in your buffer).
 --
 -- Limitations of having allow-passthrough disabled in the opencode pane:
 -- - can't display images


disrupted:
is the opencode TUI still able to receive the response and figure out the terminal capabilities if we disable tmux passthrough?

e-cal:
Sort of, `allow-passthrough` just makes tmux pass on escape sequences to the terminal instead of handling them itself. Everything still displays and works correctly, you just give up some features.

The main limitations of having it off in the opencode pane are:
- can't display images
- can't use special (terminal specific; non-system) clipboards
- may have issues setting window properties like the title

And this is all bound to only the opencode pane. Most users wouldn't notice anything different, and those that do (e.g. make use of terminal clipboards) can turn it off and use the other workaround.

e-cal:
This is why I think it should be a config option, not forced.

disrupted:
> This is why I think it should be a config option, not forced.

agreed! I've closed my initial PR since yours supersedes it

nickjvandyke:
Nice thank you, this seems like a good solution while keeping the non-focused `provider` behavior!

Could you add a bit more detail to the `disable_passthrough` docstring? Maybe suggesting to enable focus if passthrough is enabled, to still prevent the OSC code. And the features that it disables, per your above comment.

It may also be slightly easier to comprehend for users to name the provider option the same as the tmux option. i.e. `allow_passthrough` instead of its negative, `disable_passthrough`.

nickjvandyke:
Tyvm for this elegant fix! 🙏

PR #159, opened 2026-02-02
feat(server): support remote server connection and authentication

e-roux:
This PR add missing parameters to authentication as well as remove hardcoded hostname.

### DIFF
--- a/lua/opencode/cli/client.lua
+++ b/lua/opencode/cli/client.lua
@@ -3,8 +3,6 @@
 --- - [implementation](https://github.com/sst/opencode/blob/dev/packages/opencode/src/server/server.ts)
 local M = {}
 
-local config = require("opencode.config")
-
 local sse_state = {
   -- Track the port - `opencode` may have restarted, usually on a new port
   port = nil,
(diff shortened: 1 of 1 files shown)


lua/opencode/cli/client.lua
```diff
@@ -10,6 +10,39 @@ local sse_state = {
...
+  end
+
+  return nil
+end
+
+---Get the configured hostname for the opencode server.
+---@return string
+local function get_hostname()
+  local config = require("opencode.config")
+  return config.opts.hostname or "127.0.0.1"
```
nickjvandyke: Most of these defaults can be set as the default config options right? On `config.lua#defaults`. Instead of wrapping them in functions that inject defaults at runtime. Less code and easier to see 🙂

lua/opencode/cli/client.lua
```diff
@@ -3,6 +3,8 @@
 --- - [implementation](https://github.com/sst/opencode/blob/dev/packages/opencode/src/server/server.ts)
 local M = {}
 
+local config = require("opencode.config")
```
nickjvandyke: Can you inline this please? Top-level require calls hurt lazy-loading :)

nickjvandyke:
Thanks for this, very cool use case!

I assume this is for connecting to opencodes running outside your local machine right? I thought to ask you to append the hostname flag and option to `provider.cmd` like we do with port, but that probably doesn't make sense as a result.

e-roux:
I investigate the opencode server capabilities to use it on several machines.

Before I eventually make further PRs, I thought SoC would be good and wanted avoid to push this one commonly with new endpoints.

Thank you for this package!

nickjvandyke:
@e-roux did you mean to close the PR? 😅

e-roux:
I cleaned up some forks, looks like I cleaned too much.

nickjvandyke:
oops haha I figured that was the case. If it's not too much trouble, definitely open a new one! Or maybe it's possible to re-open this one if you re-fork the repo? 🤷 

I definitely want to merge this! And probably fast-follow with some config re-organization (like moving `port` and `hostname` under a new `opts.server` field.) Just don't have the time right this moment. Thanks for your patience 🙂

PR #175, opened 2026-02-13
fix(server): prevent connecting to other instances when port is explicitly configured

xyz1001:
Fixes an issue where opencode connects to an existing instance in the same directory even if a specific unique port is configured for the current nvim instance.

This change ensures that if `opts.port` is set, `opencode` will strictly attempt to connect to that port or launch a new server, skipping the discovery of other running instances via `M.select`.


nickjvandyke:
Thanks! I wasn't sure whether to retain this behavior when I added the select-server feature, but I think you're right - it should _only_ use the configured server.

PR #183, opened 2026-02-22
fix(client): add request timeout to get_path() to prevent hanging on …

zeekhuge:
## Description
This PR addresses an issue where the Neovim plugin becomes unresponsive when multiple opencode --port processes are running and one or more are stopped/unresponsive.

## Problem
When I would open quite a few opencode instances (using `opencode --port`), some of the Nvim instance would become unresponsive to the `Ctrl-a` command. 

On further investigation, I found that: the plugin discovers OpenCode servers by querying each server's /path endpoint. The curl command uses --connect-timeout 1, which only limits TCP connection time. If a process is stopped (e.g., via Ctrl+Z), the TCP socket still accepts connections but never responds, causing curl to hang indefinitely.

The current solution I use is `kill $(pidof opencode)` and then restart all of them. This works, but is a super productivity kill.
This PR is to bring in better support for the same.

## Solution
Add a `--max-time` to the `curl` commands for the `/path` query.

### DIFF
--- a/lua/opencode/cli/client.lua
+++ b/lua/opencode/cli/client.lua
@@ -41,15 +41,23 @@ local function generate_uuid()
   )
 end
 
----@param url string
----@param method string
+---@param port number
+---@param path string
+---@param method "GET"|"POST"
 ---@param body table?
----@param on_success fun(response: table)?
----@param on_error fun(code: number, msg: string?)?
+---@param on_success? fun(response: table)
+---@param on_error? fun(code: number, msg: string?)
 ---@param opts? { max_time?: number }
 ---@return number job_id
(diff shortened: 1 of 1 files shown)


zeekhuge:
Hi @nickjvandyke,
Looking forward to your feedback on this.

PS: Thank you for your amazing work on this plugin. I absolutely love using it.

nickjvandyke:
Great catch, thank you for submitting a fix! Will take a look soon. Mainly thinking about whether we want to default every request to this.

nickjvandyke:
Made a couple opinionated changes for posterity! Functionality should be unchanged and it's working for me. Let me know if you experience any issues though. Thanks again! 🙂

PR #195, opened 2026-03-03
Return only processes with ports in unix.get()

### LINKED ISSUE #194: bug: Crash due to `nil` port in `opencode.nvim` process discovery

- [x] I have read the plugin docs
- [x] I have read the [opencode docs](https://opencode.ai/docs)
- [x] I have searched the existing plugin issues

 ============================================================================== opencode:                                                                 1 ⚠️  opencode.nvim ~ - OS: Darwin 25.3.0 (arm64) - `nvim` version: `0.11.6`. - `opencode.nvim` git commit hash: `68ad02666cf050a79ae16e1b3a4df2766e0c012f`. - `vim.g.opencode_opts`: `nil`  opencode.nvim [binaries] ~ - ✅ OK `opencode` available with version `1.2.10`. - ⚠️ WARNING `opencode` version has an older `patch` version than latest tested version `1.2.11`: may cause compatibility issues.   - ADVICE:     - Update `opencode`. - ✅ OK `curl` available. - ✅ OK `pgrep` available. - ✅ OK `lsof` available.  opencode.nvim [snacks] ~ - ✅ OK `snacks.input` is enabled: `ask()` will be enhanced. - ✅ OK `snacks.picker` is enabled: `select()` will be enhanced.

### Describe the bug

*The following has been generated with opencode+gemini-3-pro-preview.*

The `opencode.nvim` plugin crashes with the error `attempt to concatenate local 'port' (a nil value)` when determining the server port. 

This occurs because the process discovery logic in `lua/opencode/cli/process/unix.lua` uses `pgrep -f "opencode .*--port"` to find candidate processes. This command matches not only the actual server process but also wrapper processes (e.g., `fish -c ...`, `node ...`) which share the command line arguments but do not listen on a TCP port.

The subsequent `lsof` check returns no port for these wrapper processes. The code then constructs a table entry `{ pid = ..., port = nil }` and returns it. When the client attempts to connect to these "servers", it tries to concatenate the `nil` port into a URL string, causing the crash.

| Software | Version |
|----------|---------|
| **OS** | macOS 26.3 (Build 25D125) |
| **Neovim** | v0.11.6 |
| **LuaJIT** | 2.1.1772148810 |
| **opencode.nvim** | 0.5.0 |
| **lsof** | 4.91 |
| **tmux** | 3.6a |
| **Ghostty** | 1.2.3 |
| **fish** | 4.5.0 |

*Note of the author:*
This issue seems to be related to #193.

### Steps To Reproduce

1. Launch `opencode` with the `--port` flag wrapped in a shell command or script (e.g., `fish -c "opencode --port ..."`).
2. Ensure `pgrep -f "opencode .*--port"` returns multiple PIDs (the wrapper and the actual process).
3. Open Neovim with `opencode.nvim` installed.
4. The plugin attempts to connect to all discovered PIDs. (e.g. executing a `require("opencode").select()`)

This results in
```
E5113: Error while calling lua chunk: .../lua/opencode/cli/client.lua:53: attempt to concatenate local 'port' (a nil value)
```

### Expected Behavior

The above steps should result in the popup of a select floating window.

*Note of the author:*
This error doesn't occur under arch-linux as of yesterday evening 2026-03-02 22:00:00 CET (more of less the same versions).

shihanng:
## Description

Prior to this change, the processes returned by `unix.get()` included all processes that matched the `pgrep` command, regardless of whether they had an associated port. This could lead to situations where processes without ports were included in the results, which were then passed to the `client.curl()` function and eventually caused errors when trying to construct a URL with a nil port value.

```
opencode.nvim/lua/opencode/cli/client.lua:53: attempt to concatenate local 'port' (a nil value)
```

This change modifies `unix.get()` to filter out processes that have no associated listening port.

Fix #194

### DIFF
--- a/lua/opencode/cli/process/unix.lua
+++ b/lua/opencode/cli/process/unix.lua
@@ -51,19 +51,15 @@ function M.get()
   -- but that misses servers started by "bun" or "node" (or who knows what else) :(
   local pgrep = vim.system({ "pgrep", "-f", "opencode .*--port" }, { text = true }):wait()
   require("opencode.util").check_system_call(pgrep, "pgrep")
-
   local pids = vim.tbl_map(function(line)
     return tonumber(line)
   end, vim.split(pgrep.stdout, "\n", { trimempty = true }))
-  local pids_to_ports = get_ports(pids)
 
+  local pids_to_ports = get_ports(pids)
   ---@type opencode.cli.process.Process[]
   local processes = {}
-  for _, pid in ipairs(pids) do
-    local port = pids_to_ports[pid]
-    if port then
-      table.insert(processes, { pid = pid, port = port })
-    end
+  for pid, port in pairs(pids_to_ports) do
+    table.insert(processes, { pid = pid, port = port })
   end
   return processes
 end


nickjvandyke:
Ty for finding a fix! I made a slight tweak to use `pids_to_ports` directly. It seems less mistake-prone to use that as the source of truth.

PR #212, opened 2026-03-23
fix(permissions): use synchronous Server instance instead of Promise

### LINKED ISSUE #211: bug: Error when OpenCode requests permission to perform an action

- [x] I have read the plugin docs
- [x] I have read the [opencode docs](https://opencode.ai/docs)
- [x] I have searched the existing plugin issues
- [x] I am using the latest version of the plugin

============================================================================== opencode:                                                                 1 ⚠️  opencode.nvim ~ - OS: Linux 6.19.9-1-cachyos (x86_64) - `nvim` version: `0.11.6+v0.11.6`. - `opencode.nvim` git commit hash: `5067c80ee1f37b5f8a54244e184df44944175a84`. - `vim.g.opencode_opts`: `nil`  opencode.nvim [binaries] ~ - ✅ OK `opencode` available with version `1.3.0`. - ⚠️ WARNING `opencode` version has an older `patch` version than latest tested version `1.2.11`: may cause compatibility issues.   - ADVICE:     - Update `opencode`. - ✅ OK `curl` available. - ✅ OK `pgrep` available. - ✅ OK `lsof` available.  opencode.nvim [snacks] ~ - ✅ OK `snacks.input` is enabled: `ask()` will be enhanced. - ✅ OK `snacks.picker` is enabled: `select()` will be enhanced.

### Describe the bug

When OpenCode needs permission to perform an action and the window doesn't have focus, a Snack.picker appears, allowing you to select what you want to do. When I select an option, it appears this issue

```
Error executing vim.schedule lua callback: ...im/lazy/opencode.nvim/plugin/events/permissions/init.lua:49: attempt to call method 'permit' (a nil value)    
stack traceback:                                                                                                                                            
        ...im/lazy/opencode.nvim/plugin/events/permissions/init.lua:49: in function 'on_choice'                                                             
        ...share/nvim/lazy/snacks.nvim/lua/snacks/picker/select.lua:59: in function <...share/nvim/lazy/snacks.nvim/lua/snacks/picker/select.lua:58>
```

I am using Astronvim with astrocommunity pack

https://github.com/AstroNvim/astrocommunity/tree/main/lua/astrocommunity/ai/opencode-nvim

### Steps To Reproduce

1. Configure opencode permissions with the "ask" option 
2. Ask Opencode something 
3. Focus other tab or window
4. Snacks.nvim will open a new prompt
5. Select 1 option
6. Boom and doesn't work

### Expected Behavior

It should not display the error, and the selected option should be sent to opencode.

saurabh-hirani:
## Summary

- Fixes `attempt to call method 'permit' (a nil value)` error when accepting/rejecting permission requests
- `Server.new(port)` was refactored to return a Promise (async validation), but the permission handlers in `plugin/events/permissions/init.lua` and `plugin/events/permissions/edits.lua` use the return value directly as a Server object
- Creates the Server instance directly with `setmetatable` since the server is already known-valid (we received an SSE event from it) — we only need the port to send the HTTP reply back
- This fix was identified with help from an LLM

Fixes #211

### DIFF
--- a/plugin/events/permissions/edits.lua
+++ b/plugin/events/permissions/edits.lua
@@ -17,7 +17,8 @@ vim.api.nvim_create_autocmd("User", {
     ---@type number
     local port = args.data.port
     ---@type opencode.server.Server
-    local server = require("opencode.server").new(port)
+    local Server = require("opencode.server")
+    local server = setmetatable({ port = port }, { __index = Server })
 
     local opts = require("opencode.config").opts.events.permissions or {}
     if not opts.enabled or not opts.edits.enabled then
--- a/plugin/events/permissions/init.lua
+++ b/plugin/events/permissions/init.lua
@@ -19,7 +19,8 @@ vim.api.nvim_create_autocmd("User", {
     ---@type number
     local port = args.data.port
     ---@type opencode.server.Server
-    local server = require("opencode.server").new(port)
+    local Server = require("opencode.server")
+    local server = setmetatable({ port = port }, { __index = Server })
 
     local opts = require("opencode.config").opts.events.permissions or {}
     if not opts.enabled then


nickjvandyke:
Thanks for your effort, but I prefer to continue using `.new` and handle the promise for consistency, rather than depend on an implementation detail 🙂

saurabh-hirani:
Got it. Makes sense. Thanks for the fix. And thank you for building opencode.nvim 👍

PR #213, opened 2026-03-24
fix: use only listed buffers for contexts

mortenhaahr:
## Description

Currently, `Context:buffers()` includes unlisted buffers, so buffers closed with :bd and opened by the LSP are still sent to the OpenCode terminal when using `@buffers`. This PR fixes that, so the context reflects the user's active working buffers.

## Related Issue(s)
No issue created

## Screenshots/Videos
I can create a video if requested


nickjvandyke:
Makes sense, ty!

PR #217, opened 2026-03-29
fix(checkhealth): check minor and patch opencode version

lungsin:
## Description

There's a minor logic error (pun intended!) in the `:checkhealth` logic when comparing the opencode version. When the minor version is larger than the latest tested version, the patched version check is still performed. So, the following cases trigger warning. (see screenshots below):
- The opencode version: 1.3.5
- The tested version: 1.2.11

## Related Issue(s)

It's probably not worth creating an issue just for this lol.

### Before

image

### After
image


nickjvandyke:
Thanks!

PR #228, opened 2026-04-08
feat(edits): support custom diff providers for edit requests

Norhua:
Closes #224
﻿
## Summary
﻿
This PR adds a configurable diff renderer for edit permission requests.
﻿
By default, `opencode.nvim` keeps using the built-in `:diffpatch` flow, but users can now provide a custom renderer through `opts.events.permissions.edits.renderer`. The renderer receives a context with the request id, filepath, unified diff, helpers to compute proposed text, send permit/reject replies, close the active view, and fall back to the default renderer.
﻿
On top of that abstraction, this PR also exposes a small `require("opencode").diff_renderers.mini_diff()` helper as an optional integration example for users who prefer a `mini.diff`-based workflow.
﻿
## What changed
﻿
- add `opts.events.permissions.edits.renderer`å
- add configurable edit keymaps under `opts.events.permissions.edits.keymaps`
- refactor edit request handling around a renderer/session abstraction
- keep the current `:diffpatch` behavior as the default renderer
- expose `require("opencode").diff_renderers`
- add `require("opencode").diff_renderers.mini_diff()` as an optional helper
- document the renderer API and `mini.diff` setup in the README
﻿
## Behavior
﻿
- existing users keep the current default behavior unless they opt into a custom renderer
- custom renderers can provide hunk navigation/actions and cleanup hooks
- global edit keymaps now operate on the active edit session and notify when no edit diff is active
- the `mini.diff` helper falls back to the default renderer if `mini.diff` is unavailable or proposed text cannot be computed
﻿
## Testing
﻿
- verified default `:diffpatch` edit flow still opens and closes correctly
- verified custom renderer path can open and manage an active edit session
- verified `mini.diff` helper can render the proposed edit and navigate hunks
- verified fallback to the default renderer when `mini.diff` is unavailable
﻿

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,7 +286,87 @@ When `opencode` requests a permission, `opencode.nvim` waits for idle to ask you
 
 #### Edits
 
-For edit requests, `opencode.nvim` opens the target file in a new tab and uses Neovim's `:diffpatch` to display the proposed changes side-by-side. See `:h 'diffopt'` for customization.
+For edit requests, `opencode.nvim` opens the target file in a new tab and displays the proposed changes with its built-in `:diffpatch` renderer by default. See `:h 'diffopt'` for customization.
+
+You can replace the renderer with a custom function. The renderer receives a context and can return a session with optional hunk navigation, hunk actions, and cleanup hooks.
+
+The context provides:
+
+- `ctx.request_id`: the edit request id.
+- `ctx.filepath`: normalized target filepath.
+- `ctx.diff`: the unified diff from `opencode`.
+- `ctx.proposed_text()`: lazily computes the patched file contents.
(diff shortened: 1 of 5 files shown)


nickjvandyke:
Please respond to the automated review comments and make sure it passes the checks 🙂

Norhua:
After a few days of testing, I found that existing diff plugins struggle to render Unified Diffs well due to their 'change' semantics. In this case, it's probably better for me to write a simple diff tool that only uses 'add' and 'delete' semantics. Because of this, I'm closing this PR.

nickjvandyke:
Thanks for experimenting! Feel free to let me know anything else useful you discover 🙂

PR #234, opened 2026-04-15
allow terminal setup to preserve backend focus

nkabbara:
Hey @nickjvandyke!

I started customizing the plugin to open in a float as I am improving my workflow with ai. I noticed that I could not set the cursor to stay in my newly opened float because of the `setup` method. Do you mind if we made it configurable to do so?

## Summary
- add an optional `restore_focus` flag to `opencode.terminal.setup()` while keeping the current behavior as the default
- let custom terminal integrations preserve their own focus behavior by calling `setup(win, { restore_focus = false })`
- update the README `snacks.terminal` example to show how to opt out of the forced focus restore

## Why
Custom terminal backends may want to control whether focus stays in the terminal or returns to the previous window. The current `setup()` helper always restores focus as part of its redraw workaround, which makes it hard for custom integrations to keep the terminal focused.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,8 +157,9 @@ local snacks_terminal_opts = {
     position = 'right',
     enter = false,
     on_win = function(win)
-      -- Set up keymaps and cleanup for an arbitrary terminal
-      require('opencode.terminal').setup(win.win)
+      -- Set up keymaps and cleanup for an arbitrary terminal.
+      -- `restore_focus = false` preserves the terminal backend's focus choice.
+      require('opencode.terminal').setup(win.win, { restore_focus = false })
     end,
   },
 }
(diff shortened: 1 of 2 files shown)


nickjvandyke:
Thanks for calling this out! I pushed a change just now that moves the redraw workaround to only our terminal implementation, because it seems unnecessary for `snacks.terminal`. Hopefully that holds true for other terminal integrations too? Let me know if that works for you.

nkabbara:
> Thanks for calling this out! I pushed a change just now that moves the redraw workaround to only our terminal implementation, because it seems unnecessary for `snacks.terminal`. Hopefully that holds true for other terminal integrations too? Let me know if that works for you.

Just saw this. Trying.

nkabbara:
> Thanks for calling this out! I pushed a change just now that moves the redraw workaround to only our terminal implementation, because it seems unnecessary for `snacks.terminal`. Hopefully that holds true for other terminal integrations too? Let me know if that works for you.

Works great! Thanks!

PR #236, opened 2026-04-17
feat: add tab-aware terminal/server context

nkabbara:
Summary
- Make terminal state tab-scoped (toggle/open/close)
- Make connected server + SSE subscription tab-scoped
- Use tab-local context in Server.get and related UI callsites

### DIFF
--- a/lua/opencode/events.lua
+++ b/lua/opencode/events.lua
@@ -92,9 +92,13 @@ function M.connect(server, tab)
         refresh_compat_connected_server()
 
         if state.heartbeat_timer then
-          state.heartbeat_timer:start(OPENCODE_HEARTBEAT_INTERVAL_MS + 5000, 0, vim.schedule_wrap(function()
-            M.disconnect(tab)
-          end))
+          state.heartbeat_timer:start(
+            OPENCODE_HEARTBEAT_INTERVAL_MS + 5000,
+            0,
+            vim.schedule_wrap(function()
+              M.disconnect(tab)
+            end)
+          )
         end
 
         if require("opencode.config").opts.events.enabled then
--- a/lua/opencode/events.lua
+++ b/lua/opencode/events.lua
@@ -11,11 +11,59 @@ local M = {}
 ---
 ---@field permissions? opencode.events.permissions.Opts
 
-local heartbeat_timer = vim.uv.new_timer()
 ---How often `opencode` sends heartbeat events.
 local OPENCODE_HEARTBEAT_INTERVAL_MS = 30000
----@type number?
-local subscription_job_id = nil
+
+---@class opencode.events.State
+---@field heartbeat_timer uv_timer_t
+---@field subscription_job_id? number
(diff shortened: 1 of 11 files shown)


nkabbara:
Hey @nickjvandyke!

I'm updating my workflow to use git worktrees where each tab in nvim is bound to a worktree. I was having issues with the plugin because of tab interference. ie. tabs are many to one (opencod). This scopes opencode terms/instances to different tabs.

Do you forrsee merging this (or something similar/better)?

nickjvandyke:
Hmm this is a great point. I have thought about supporting multiple simultaneous connections, at least for SSEs. But it introduces complications like how to select a server when prompting. By tab is a neat idea, but we need a more general solution. (or maybe not? if this only realistically happens in multi-tab workflows)

nkabbara:
> Hmm this is a great point. I have thought about supporting multiple simultaneous connections, at least for SSEs. But it introduces complications like how to select a server when prompting. By tab is a neat idea, but we need a more general solution. (or maybe not? if this only realistically happens in multi-tab workflows)

Yeah this is totally born out of a need of mine. But if someone still wants to control the same session, with this solution, they can still select it using `server.select`. Though,  I think tabs in vim should be thought of as workspaces . E.G https://stackoverflow.com/questions/26708822/why-do-vim-experts-prefer-buffers-over-tabs 

In the meantime I can work off of my fork until a decision is made.

nickjvandyke:
Sorry for all the merge conflicts! Hopefully the refactors simplify your work on top of it though.

(Still thinking on this, so no urgent need to resolve them - but in case you wanted to for the sake of your fork)

nkabbara:
> Sorry for all the merge conflicts! Hopefully the refactors simplify your work on top of it though.
> 
> (Still thinking on this, so no urgent need to resolve them - but in case you wanted to for the sake of your fork)

Hi @nickjvandyke!

I totally forgot that I was running off of my forked branch all this time! I must have missed some cool features. The setup has been working great for me. I work on multiple features simultaneously in separate tabs and being able to easily spin up different opencodes in isolation has become a requirement for me.

Will rebase and continue working from here in the meantime.

nkabbara:
At first glance all seem to be good after rebase. Heavy coding tomorrow, so I should catch issues if they come up.

nickjvandyke:
@nkabbara I recently removed `terminal.lua` for maintenance reasons, preferring users to use a dedicated terminal plugin for complex needs. Do you know if your changes work with terminal plugins and/or simply `vsplit term://opencode --port` in the tabs? I don't see why not - seems like your `terminal.lua` changes just added tab-handling to the buffers and windows it manages (which is why I removed it altogether before it grows into a terminal plugin of its own via general scope creep lol).

nkabbara:
Hey @nickjvandyke! Thanks for the update. I am using this less and less these days and I think when opencode v2 comes out that my whole flow will change. I don't think I'll need this anymore.

Issue #190, opened 2026-02-27
feature: interact with `opencode` server without TUI


marcelbeumer:
### Did you check the docs?

- [x] I have read all the docs

### Is your feature request related to a problem? Please describe.

I like to run my opencode in a sandbox with `opencode serve`. Maybe on `http://127.0.0.1:4069`, maybe elsewhere. I notice the plugin tries to be smart about running servers, but right now to make this work I need to do:

1. Start opencode with `opencode serve --port` because without `--port` the plugin won't detect the process
2. Add a configuration block for `require("opencode.terminal").start|toggle` on with `("opencode attach http://127.0.0.1:4096")` instea of `("opencode --port")`

And I guess it won't work with opencode instances on other hosts at all. 

Why not just provide and endpoint via the config, and when its empty, try to start a own server (or not; i can start my own)? No smart process finding, etc.

### Describe the solution you'd like

Provide and endpoint via the config, and when its empty, try to start a own server (or not; i can start my own)? No smart process finding, etc.

### Describe alternatives you've considered

Not much

nickjvandyke:
> without --port the plugin won't detect the process

Ah, with just `opencode`, `--port` is required to expose the server. I didn't realize that `serve` does that too! Changing the detection to pick that up should be easy.

I think that would address 2. as well? `start` is irrelevant when a server is found.

I'd definitely like to make the server location more flexible. Probably via `opts.server.origin`? I just need time to consider how it interplays with the existing server discovery and starting logic.

nickjvandyke:
Does `opencode serve` even work with requests from the plugin? I see it receive requests with `--print-logs`, but it doesn't actually do anything (e.g. add requested comments)

nickjvandyke:
> And I guess it won't work with opencode instances on other hosts at all.

I overlooked this initially. You mean with no TUI at all, right? That'd require us to use a different (and more difficult) API entirely to send directly to the server. It also re-introduces the problem that we can't know the "active" session, and have to best-guess it as the most recently interacted one. I guess I'll have to see how `opencode` usage evolves.

But I think we can at least accommodate sending prompts to a TUI attached to a remote server? It depends. I wonder if we are meant to send TUI requests to the server still, or if the attached TUI process exposes its own server.

nickjvandyke:
@linjiang82 I noticed you implemented remote server support in your fork! Does it work? Would you be interested in opening a PR? 🙂

thecacho:
[...]

PR #239, opened 2026-04-21
fix(server): send x-opencode-directory: nvim-cwd on every request

juanperi:
## Tasks

- [x] I filled out this PR template
- [x] I responded to and/or resolved automated PR comments

### The bug

In setups where Neovim talks to a shared `opencode serve` that also has one or more `opencode attach --dir <path>` TUIs, `<leader>oq` / `opencode.prompt` / `opencode.ask` / the `tui.command.execute` commands appear to succeed (HTTP 200, no errors) but the prompt never appears in any TUI.

It's silent: no error, no notification, no visible effect.

### Root cause

The opencode TUI's event subscription filters events by the request's _instance directory_:

```ts
// packages/opencode/src/cli/cmd/tui/context/event.ts (~line 28, v1.4.11)
if (event.directory === project.instance.directory()) {
  handler(event.payload)
}
```

Each HTTP request to the opencode server establishes an Instance scope via either the `x-opencode-directory` header or a `?directory=` query parameter. `tui.*` events published by that request inherit the instance directory.

Previously, `Server:curl` in this plugin didn't send that header. As a result, every `tui.prompt.append` event carried the server's ambient cwd, which doesn't match any attach-TUI's `--dir`. The TUI subscriptions filtered the events out → silent no-op.

### The fix

Add `x-opencode-directory: <nvim cwd>` to every HTTP request in `Server:curl`.

### Compatibility

- **Single-TUI / integrated-mode setups (most users today):** no behavior change. The only TUI is the one nvim launches; its `--dir` is Neovim's cwd; the header matches trivially.
- **Multi-TUI setups (`opencode serve` + `opencode attach`):** previously broken, now correct.
- **Zero config changes required.** The fix is automatic.

### Testing

End-to-end reproduction on a real multi-TUI setup (opencode v1.4.11):

1. `opencode serve --port 4096` in the background.
2. `opencode attach http://127.0.0.1:4096 --dir ~/` in tmux pane A.
3. `opencode attach http://127.0.0.1:4096 --dir ~/somewhere` in tmux pane B.
4. Open Neovim in `~/somewhere`. Config: `vim.g.opencode_opts = { server = { port = 4096, start = false } }`.
5. Fire `:lua require("opencode").prompt("marker text")`.

> Note: `~/somewhere` must exist otherwise opencode will not start

**Before this PR:** HTTP 200 from `/tui/publish`, server log shows `type=tui.prompt.append publishing`, neither TUI shows anything.

**After this PR:** TUI A's prompt input shows `marker text`. TUI B is untouched.

## Related Issue(s)

Partially addresses #190. That issue asks for a way to make the plugin work with `opencode serve` + remote `opencode attach` setups, and captures the maintainer's concern: _"that'd require us to use a different (and more difficult) API entirely to send directly to the server. It also re-introduces the problem that we can't know the 'active' session, and have to best-guess it as the most recently interacted one."_

This PR resolves that concern without introducing any new API and without having to guess the active session: nvim's `getcwd()` is a natural, deterministic disambiguator that opencode's server already honors. It keeps the plugin's current TUI-based design intact (prompts still land in the TUI's input box; the user can still edit/resubmit) — the only thing that changes is that the right TUI receives them.

Not a complete solution to #190 on its own (there's also the discussion of remote hosts, password auth, message persistence, etc.), but it unblocks the core multi-TUI local case and should compose cleanly with whatever remote-server solution eventually lands.

### DIFF
--- a/lua/opencode/server/init.lua
+++ b/lua/opencode/server/init.lua
@@ -98,6 +98,16 @@ function Server:curl(path, method, body, on_success, on_error, opts)
     "-N",
   }
 
+  -- Route the request to the TUI whose working directory matches Neovim's.
+  -- In single-TUI / integrated-mode setups this matches the only TUI trivially.
+  -- In multi-TUI setups (`opencode serve` + several `opencode attach --dir ...`)
+  -- this ensures `tui.*` events are delivered to the intended TUI.
+  local nvim_cwd = vim.fn.getcwd()
+  if nvim_cwd and nvim_cwd ~= "" then
+    table.insert(command, "-H")
+    table.insert(command, "x-opencode-directory: " .. nvim_cwd)
+  end
+
   if opts.max_time then
     table.insert(command, "--max-time")
     table.insert(command, tostring(opts.max_time))
--- a/lua/opencode/server/init.lua
+++ b/lua/opencode/server/init.lua
@@ -137,6 +137,16 @@ function Server:curl(path, method, body, on_success, on_error, opts)
     table.insert(cmd, 2)
   end
 
+  -- Route the request to the TUI whose working directory matches Neovim's.
+  -- In single-TUI / integrated-mode setups this matches the only TUI trivially.
+  -- In multi-TUI setups (`opencode serve` + several `opencode attach --dir ...`)
+  -- this ensures `tui.*` events are delivered to the intended TUI.
+  local nvim_cwd = vim.fn.getcwd()
+  if nvim_cwd and nvim_cwd ~= "" then
+    table.insert(cmd, "-H")
+    table.insert(cmd, "x-opencode-directory: " .. nvim_cwd)
+  end
+
   if body then
     table.insert(cmd, "-d")
     table.insert(cmd, vim.fn.json_encode(body))


nickjvandyke:
Oo, thank you for this fix and explanation! That clears up a lot for me re #190.

Unfortunately this seems to break when we run a single `opencode` in a different directory than Neovim. That used to work, and now the TUI command is dropped.

I guess when only one TUI is attached to the server, the server routes the request to it without needing `x-opencode-directory`? But now that we've added the header, that check fails in the single-TUI scenario?

juanperi:
> Unfortunately this seems to break when we run a single `opencode` in a different directory than Neovim. That used to work, and now the TUI command is dropped.

ohh, that is something that I've never done 😓 , i always run nvim and the opencode in the same folder.
I think what's happening in that case is that when you do not send the header, the commands land in whatever pwd the server has, and because it's different than the one we send, it is completely ignored

I honestly do not know what you would like to do in those scenarios. Do we need to introduce configuration? (painnnn) or request the tui in the same folder?

If it is a configuration, it could be like:

```
vim.g.opencode_opts = {
  server = {
    port = 4096,
    start = false,
    directory = vim.fn.getcwd(),  -- or a static path, or a function
  },
}
```
Three modes fall out of a single field:

* nil / unset (default): No header sent. Exact pre-PR behavior. Single-TUI setups just work again.
* a string path (e.g. "/Users/foo/projB"): Send x-opencode-directory: /Users/foo/projB on every request → routes to the TUI attached with --dir /Users/foo/projB.
* "cwd" (magic string) or a function: Resolve at request time to nvim's current working directory → matches the TUI attached to nvim's project. The current PR's behavior, but opt-in.

Not sure if i'm overthinking it. I honestly thought it was a "transparent" fix. But something like this would force you to change the surface area of your plugin

nickjvandyke:
> i always run nvim and the opencode in the same folder.

I do too, and I bet that's by far the most common. But the plugin allows connecting to _any_ server because idk everyone's workflows, and there are some edge cases. e.g. see #104.

Obviously completely different CWDs are unusual and less important to support. But this also affects CWDs that partially overlap. Imagine a directory containing inter-related projects - a user might run `opencode` at the root (say `~/dev`), and then nvim inside the currently most relevant project (say `~/dev/opencode.nvim`).

Thanks for detailing some options! You're not overthinking it - decisions like this just have many considerations 😅 I agree additional configuration is not ideal. Atm I don't see a way to reliably internalize this though 🤔

I will chew on this when I have more time. Definitely open to any other thoughts that cross your mind too.

juanperi:
In light of this twit, this PR might not be needed anymore
https://x.com/thdxr/status/2071007637487894883?s=20

> in OpenCode v2 all instances of the tui and desktop and web share the same backend
> so everything is synced by default and resource usaged is minimized no matter how many windows you have open

nickjvandyke:
Merging this to prep for OpenCode v2 changes! Similar to v1, its server uses this header to route to the matching client (TUI). But all local clients now share a single local daemon server. Thus the "fallback to the only attached client" routing logic won't apply nearly as often.

And, some brief testing shows that this fancy dancy new daemon server walks up directories _for_ us! We no longer need to do that ourselves - just send our CWD and it'll find the likely match. So the "exact match" issue that blocked this PR is solved! 😁

PR #242, opened 2026-04-27
fix(server): emit one process entry per (PID, port) pair

erroris3:
## Tasks

- [x] I filled out this PR template
- [x] I responded to and/or resolved automated PR comments

## Description

A single `opencode` PID can listen on multiple ports (e.g. when its TUI and HTTP server bind separately, or when multiple listeners are opened during startup). The previous `get_ports` implementation in `lua/opencode/server/process/unix.lua` collected results into a `pids_to_ports` map keyed by PID, so each new port for the same PID overwrote the previous one. Only the last port emitted by `lsof` survived, and any earlier listening ports for that PID were silently dropped from server discovery.

This change replaces the PID-keyed map with a flat list of `{ pid, port }` entries, appending one entry per port reported by `lsof`. Every listening port now surfaces to the discovery layer, which already deals in `opencode.server.process.Process[]`.

While here:

- Renamed `get_ports` to `get_processes_with_ports` to reflect the new return shape.
- Added a `nil` guard on the parsed port so malformed `lsof` lines do not insert a `{ pid, port = nil }` entry.
- Removed the now-redundant map-to-list conversion in `M.get`, which just forwards the helper's return value.

No public API changes; the function is local to the module and `M.get` keeps its existing return type.

## Related Issue(s)
image

### DIFF
--- a/lua/opencode/server/process/unix.lua
+++ b/lua/opencode/server/process/unix.lua
@@ -1,9 +1,9 @@
 local M = {}
 
 ---@param pids number[]
----@return table<number, number>
-local function get_ports(pids)
-  assert(#pids > 0, "`get_ports` should only be called with a non-empty list of PIDs to filter by")
+---@return opencode.server.process.Process[]
+local function get_processes_with_ports(pids)
+  assert(#pids > 0, "`get_processes` should only be called with a non-empty list of PIDs to filter by")
 
   local lsof = vim
     .system({
@@ -23,7 +23,8 @@ local function get_ports(pids)
     :wait()
   require("opencode.util").check_system_call(lsof, "lsof")
 
-  local pids_to_ports = {}
+  ---@type opencode.server.process.Process[]
+  local processes = {}
   local pid
   for line in lsof.stdout:gmatch("[^\n]+") do
     local prefix = line:sub(1, 1)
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Nice find, thank you for a fix!

> e.g. when its TUI and HTTP server bind separately

Is it possible to change the `pgrep` or `lsof` command to only capture the server? We don't send requests to the TUI. I assume not, and the later `Server.new` checks should filter out the TUI. But would be nice to pre-empt that.

erroris3:
> Nice find, thank you for a fix!
> 
> 
> 
> > e.g. when its TUI and HTTP server bind separately
> 
> 
> 
> Is it possible to change the `pgrep` or `lsof` command to only capture the server? We don't send requests to the TUI. I assume not, and the later `Server.new` checks should filter out the TUI. But would be nice to pre-empt that.


Sorry, that was my mistake. I’m not very good at English, so I used AI to generate the description and didn’t check it carefully.

The open port is not related to the TUI, and the current command that you did is already correct.

I found that some opencode plugins can open additional ports for example something like CLIProxyAPI.

So the port being detected is likely coming from one of those plugins, nothing relate to the TUI itself.

PR #245, opened 2026-05-01
fix: process discovery on NixOS

ffloyd:
## Tasks

- [ ] I responded to and/or addressed automated review comments (if any)

## Description

On NixOS `ps -au` can return the following line for OpenCode:

```
ffloyd    326815  5.4  1.2 74522096 411196 pts/2 Ssl+ 16:16   0:11 /nix/store/aqdisqwd5bxp8szz9nrg9sz0d24knmnp-opencode-adjusted/bin/.opencode-wrapped_ --port
```

It happens because Nix often use wrappers. The pgrep pattern should be adjusted. Otherwise - no OpenCode process can be found by the plugin.

Theoretically, it can lead to false-positives, but I suspect it's not a big problem.

### DIFF
--- a/lua/opencode/server/process/unix.lua
+++ b/lua/opencode/server/process/unix.lua
@@ -54,7 +54,9 @@ function M.get()
   -- Filter by `--port` because it's required to expose the server.
   -- We can aaaalmost skip this and just use "-c opencode" with `lsof`,
   -- but that misses servers started by "bun" or "node" (or who knows what else) :(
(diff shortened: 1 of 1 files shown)


nickjvandyke:
Thanks! It seems I added a space in https://github.com/nickjvandyke/opencode.nvim/commit/fefc9cfb67454b3d6f2594417dc78a282c3b4e57 but not for any particular reason, so seems safe to add it back.

PR #252, opened 2026-05-05
fix: narrow pgrep regex so wrappers (proxy-tool, npm/npx, node) are n…

dfavato:
## Tasks

- [X] I read [CONTRIBUTING.md](https://github.com/nickjvandyke/opencode.nvim/blob/main/CONTRIBUTING.md)
- [X] I reviewed AI-generated code myself
- [X] I made sure my changes pass automated checks
- [ ] I responded to and/or addressed automated review comments (if any)

## Description

The previous regex 'opencode.*--port' matched any process whose full command line contained those tokens in order. In environments where opencode is launched through wrappers (e.g. 'proxy-tool exec -- npx -y opencode-ai@latest --port ...'), wrapper processes such as proxy-tool, npm, and the node launcher all match. Some of them also listen on a TCP port (proxy-tool runs an HTTP forward proxy), so the subsequent lsof + curl /path probing hit them, producing spurious 'Response decode error' notifications and visible errors in the embedded TUI (e.g. 'ERROR proxy_tool > GET /path - unsupported').

Anchor the regex to argv[0] and require it to end in 'opencode' so only the real opencode binary matches. Also route JSON-decode failures through on_error when the caller provides one, so speculative server discovery doesn't pop a global notification on every non-opencode port that happens to respond.

### DIFF
--- a/lua/opencode/cli/server.lua
+++ b/lua/opencode/cli/server.lua
@@ -20,8 +20,12 @@ end
 ---@return opencode.cli.server.Process[]
 local function get_processes_unix()
   -- Find PIDs by command line pattern.
-  -- We filter for `--port` to avoid matching other `opencode`-related processes (LSPs etc.)
-  local pgrep = vim.system({ "pgrep", "-f", "opencode.*--port" }, { text = true }):wait()
+  -- We filter for `--port` to avoid matching other `opencode`-related processes (LSPs etc.).
+  -- The leading `[^ ]*` constrains the match to processes whose argv[0] ends in `opencode`,
+  -- so wrappers/launchers that happen to mention `opencode` somewhere later in their argv
+  -- (e.g. `proxy-tool exec -- npx -y opencode-ai@latest --port`, `npm exec opencode-ai@latest`,
+  -- `node .../opencode-ai/bin/.opencode --port`) don't get picked up.
+  local pgrep = vim.system({ "pgrep", "-f", "^[^ ]*opencode --port" }, { text = true }):wait()
   require("opencode.util").check_system_call(pgrep, "pgrep")
 
   local processes = {}
--- a/lua/opencode/cli/client.lua
+++ b/lua/opencode/cli/client.lua
@@ -83,6 +83,10 @@ local function curl(url, method, body, on_success, on_error)
           if on_success then
             on_success(response)
           end
+        elseif on_error then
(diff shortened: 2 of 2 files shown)


dfavato:
Sorry, I was working on my exact commit.. closing this

nickjvandyke:
Ope, I guess this is why I originally added the space that was just removed in https://github.com/nickjvandyke/opencode.nvim/pull/245 😅 I couldn't remember or find any notes in the first commit.

> Also route JSON-decode failures through on_error when the caller provides one, so speculative server discovery doesn't pop a global notification on every non-opencode port that happens to respond.

This is a fix by itself right? Sure it's a bit of extra work, but negligible imo - on subsequent calls, `opencode.nvim` uses the connected server and doesn't re-discover. iirc in https://github.com/nickjvandyke/opencode.nvim/pull/244 I assumed that curl would error and we'd skip JSON decoding entirely (thus not needing this additional `on_error` check), but seems I was wrong!

> Sorry, I was working on my exact commit.. closing this

Not quite sure what you mean - do you plan to re-open a different PR, or should I make these changes myself?

Thank you!

lua/opencode/cli/server.lua
```diff
@@ -0,0 +1,290 @@
...
+end
+
+---@return opencode.cli.server.Process[]
+local function get_processes_unix()
+  -- Find PIDs by command line pattern.
+  -- We filter for `--port` to avoid matching other `opencode`-related processes (LSPs etc.).
+  -- The leading `[^ ]*` constrains the match to processes whose argv[0] ends in `opencode`,
+  -- so wrappers/launchers that happen to mention `opencode` somewhere later in their argv
+  -- (e.g. `proxy-tool exec -- npx -y opencode-ai@latest --port`, `npm exec opencode-ai@latest`,
+  -- `node .../opencode-ai/bin/.opencode --port`) don't get picked up.
```
nickjvandyke, line 27: Sometimes we want to find these processes, e.g. when installed via `bun`, the desired process looks like `bun [...] opencode`

nickjvandyke:
> > Sorry, I was working on my exact commit.. closing this
>
> Not quite sure what you mean - do you plan to re-open a different PR, or should I make these changes myself?

Ah I see now that this targets a really old version of `main`. Try the latest `main`! #245 adding `--fail-with-body` to `curl` maaaay resolve the global error notifications. If not, then we need the `on_error` branch you added to `process_response_buffer`.

PR #263, opened 2026-05-14
fix: send interrupt command immediately

### LINKED ISSUE #257: bug: `session.interrupt` command requires two sends
### Describe the bug

When we send the `session.interrupt` command, the `opencode` TUI displays "press Esc again to interrupt".

I think this regressed with `opencode` v1.0 because it used to work immediately.

### Steps To Reproduce

1. Send a message to `opencode` so it begins responding
2. Select `session.interrupt` command from `opencode.nvim`
3. Must repeat step 2 to actually interrupt.

### Expected Behavior

Selecting `session.interrupt` command a single time interrupts `opencode`.

Normally I don't like to work around `opencode` bugs, but this one is simple.

See `lua/opencode/api/command.lua`

mturac:
## Summary
- send `session.interrupt` twice from the Neovim command wrapper so one user action interrupts opencode v1.0 responses
- keep all other commands unchanged

Fixes #257

## Checks
- `luac -p lua/opencode/api/command.lua`
- `git diff --check`

### DIFF
--- a/lua/opencode/api/command.lua
+++ b/lua/opencode/api/command.lua
@@ -26,6 +26,10 @@ local M = {}
 function M.command(command)
   return require("opencode.server").get():next(function(server) ---@param server opencode.server.Server
     server:tui_execute_command(command)
+    -- opencode v1.0 prompts for a second interrupt; keep the Neovim command single-action.
+    if command == "session.interrupt" then
+      server:tui_execute_command(command)
+    end
   end)
 end
 


nickjvandyke:
Ignoring the PR template (and [this](https://github.com/mvanhorn/cli-printing-press/pull/1381#issuecomment-4452000028)) signals to me that this is a drive-by AI PR, which defeats the purpose of "good first issues". Closing.

PR #287, opened 2026-06-07
Fix E816 malformed diffpatch output issue

zawedcvg:
## Tasks

- [ ] I addressed actionable AI review feedback (if any)

## Description

Issue basically was because of some incorrect file path handling. The current solution does work, but might not be the best, so let me know. Below is what the logs captured

```
2026-06-06 22:39:41 [opencode edits] opened edit diff buffer
{
  buf_name = "/home/user/.config/nvim/.config/nvim/lua/plugins/opencode.nvim/lua/opencode/events/permissions/edits.lua",
  buftype = "",
  current_buf = 38,
  current_tabpage = 2,
  current_win = 1038,
  error = "vim/_core/editor.lua:0: nvim_exec2(), line 1: Vim(diffpatch):E816: Cannot read patch output",
  filetype = "lua",
  modified = false,
  ok = false,
  windows_in_tab = { {
      buf = 38,
      diff = false,
      name = "/home/user/.config/nvim/.config/nvim/lua/plugins/opencode.nvim/lua/opencode/events/permissions/edits.lua",
      win = 1038
    } }
}
```
## Testing

1. 
2. 
3.


nickjvandyke:
Thanks for taking a crack at this!

I'm a bit confused - why would this affect the `diffpatch` error? `filepath` is only used to open the current version of the edited file in `tabnew`. Whereas `diffpatch` receives `patch_filepath`, untouched by this PR.

zawedcvg:
I could be wrong but I think its due to this from the ```:help diffpatch```

Note that {patchfile} should only contain a diff for one file,
		the current file.  If {patchfile} contains diffs for other
		files as well, the results are unpredictable.
		
I think because its trying to diff with a file that doesn't exist, the error is being thrown. At least this is my guess for now.

zawedcvg:
Ok you can reproduce the E816 pretty easily. have a file and generate a random patch for it and store it somewhere. Then open a non existing file and diffpatch with this patch. You get E816.

Also interestingly, if you open diffpatch from an existing file, but the file is not the correct one, you get this, which is a thing that sometimes occurs with opencode
image

nickjvandyke:
Ah I see, thanks for explaining! So basically OpenCode might send a filepath that's incorrect relative to how Neovim searches for the file to open it? And resolving it as an absolute path fixes that?

Why do we sometimes make it relative to the home directory? I don't see when the absolute path would fail and need to fallback to that.

> Also interestingly, if you open diffpatch from an existing file, but the file is not the correct one, you get this, which is a thing that sometimes occurs with opencode

Ah thanks for figuring that out! It puzzled me. Does this change fix it too?

nickjvandyke:
Given we have the `server` available in this scope, I wonder if `local filepath = server.cwd .. event.properties.metadata.diff` would be even more reliable? I presume OpenCode always sends the filepath relative to its CWD. Whereas Neovim could very rarely resolve the absolute path differently with very particular directory structures and CWD mismatches.

zawedcvg:
from what I remember the home directory branch was added because just making it absolute didn't completely solve. I might be misremembering but 
> local filepath = server.cwd .. event.properties.metadata.diff

doesn't solve it either. It has the same path issue because open code doesn't create paths wrt server.cwd. 

>Ah thanks for figuring that out! It puzzled me. Does this change fix it too?

Nope, it doesn't, just the malformed issue. I will spend sometime to see if I can find a better fix to the malformed issue and also fix the single line issue. Though the latter is a bit confusing in the sense that it opens the correct file, but the patch is not for that file? It doesn't make complete sense to me but I suspect it to be a similar file path issue. Sucks that diff patch doesn't have the greatest error handling but what can you do.

zawedcvg:
for the e816 error, I think this is the fix for now. Opencode sometimes sends the full path and sometimes everything but home, so I dont think there is a better solution.

I managed to figure out the bug with the single line thing, bad news is that there's nothing that can be done about it. So turns out that what opencode exposes in the diff, is a pretty printed version of the actual diff. It uses it just for display. It calls ```trimDiff``` on the actual unified patch, which can cause the patch to be malformed. Basically anytime the edit diff actually works with the plugin, it is due to luck. So yeah nothing can be done about this from the plugin side. I will open an issue and see if the opencode devs want to expose the actual patch, since it doesn't serve any purpose except for niche usecases like this.

zawedcvg:
addressed the comments

lua/opencode/events/permissions/edits.lua
```diff
@@ -20,6 +20,22 @@ function M.diff(event, server)
...
+    -- Opencode sends the absolute path sometimes with the HOME and sometimes without
+    -- It has something to do with the path of the opencode server cwd wrt the file/directory
+    if vim.fn.filereadable(absolute_filepath) == 1 then
+      filepath = absolute_filepath
+    elseif vim.env.HOME and vim.env.HOME ~= "" then
+      local home_filepath = vim.fs.normalize(vim.fs.joinpath(vim.env.HOME, filepath))
+      if vim.fn.filereadable(home_filepath) == 1 then
+        filepath = home_filepath
+      end
+    else
```
nickjvandyke: It's unlikely, but I see a small bug here: the `home_filepath` readability check is in a nested `if`, so if it fails, we'll miss the outer `else` w/ the error notification.

Feel free to restructure how you like, but maybe after the absolute/home resolution attempts to update `filepath`, we put a standalone `vim.fn.filereadable(filepath)` check? Then no matter the code path, we check the final `filepath` before attempting to use it.

I think we also want an early `return` after the notification, so we don't get E816. And maybe `vim.log.levels.ERROR`.

lua/opencode/events/permissions/edits.lua
```diff
@@ -84,6 +100,7 @@ function M.diff(event, server)
     end, { buffer = true, desc = "Close opencode edit diff" })
   elseif event.type == "permission.replied" and current_edit_request_id == event.properties.requestID then
     -- Entire edit was accepted or rejected, either in the plugin or TUI; close the diff
+    -- #NOTE: i don't think it works when accepting/rejecting from the TUI
```
nickjvandyke: This worked last I tested, has that changed? Feel free to open a new bug report if so, but let's keep this PR focused 🙂

zawedcvg: oops, wasnt supposed to be part of the commit. i seemed to have some .new files throughout the directory when bug fixing this, but yeah, will take a look and create a different bug report if so.

nickjvandyke:
Thank you for investigating this tricky issue!

PR #304, opened 2026-06-24
Feature: Un-intrusive floating popup for opencode permissions 

r0ckHopper:
## Tasks

- [X] I read [CONTRIBUTING.md](https://github.com/nickjvandyke/opencode.nvim/blob/main/CONTRIBUTING.md)
- [X] I ensured my changes pass automated checks

### Issue 1 and Solution
The current 2 options (vim.ui.select and snacks) for permissions from opencode break the flow of coding, I found that while I was doing edits, opencode kept giving me popups and I wasn't ready to context switch yet. So I needed a non-intrusive solution to reviewing opencode requests. 

This plugin addition is an optional floating window for Permissions that is not focused by default. 
It keeps a queue of all concurrent requests that opencode made which can be navigated through and approved one at a time once user is ready to do so. 

### Issue 2 and Solution
if I hit cancel on the requests I couldn't go back to viewing them from within nvim which meant I had to use the tui to accept

This change allows user to toggle hide for permissions and go back to work, then toggle it back without losing state or needing to use tui. 

### Issue 3 and Solution
sometimes permissions can get out of sync if user replies in the tui. So nvim still shows the permission even though it is no longer up

Opencode server already sends events notifying us of this, so we can update when we receive this request.

### Other niceties added
- Allow All option when multiple requests stack
  - does it sequentially. I was careful to avoid race condition because I saw that if the requests were done out of order they would fail spectacularly
- Auto update buttons
  - When multiple requests enter queue, popup re-renders to update buttons and number of permissions
  - Reject is updated to reject all (Opencode rejects all requests when you send reject, this was a gap in original vim.ui.select implementation of permissions, once you reject 1st request you cannot accept next one)
- Multiple lines are formatted correctly so we can see better
- Update highlight of border based on focus 
- Added some permission labels like $ for bash commands (just 4 for now)

### Some design decisions
- Closing the window hides it instead so u can still toggle back, only way to get rid of this is to reject
- Diff ui is unaffected, you will not see this floating pop up when opencode requests an edit (I think current da and dr works well)

### How to configure 
This is an optional feature and requires user to add the following to opts
```lua
vim.g.opencode_opts = {
	events = {
		permissions = {
			floating = true,
		},
	},
}
```
Additionally they can keybind to most functions by doing the following
```lua
vim.keymap.set({ "n", "v", "i" }, "<C-h>", function() require("opencode.events.permissions.floating").move_left() end, { desc = "Permission: select left" })
vim.keymap.set({ "n", "v", "i" }, "<C-l>", function() require("opencode.events.permissions.floating").move_right() end, { desc = "Permission: select right" })
vim.keymap.set({ "n", "v", "i" }, "<C-n>", function() require("opencode.events.permissions.floating").next_permission() end, { desc = "Permission: next request" })
vim.keymap.set({ "n", "v", "i" }, "<C-p>", function() require("opencode.events.permissions.floating").prev_permission() end, { desc = "Permission: prev request" })
vim.keymap.set("n", "<leader><CR>", function() require("opencode.events.permissions.floating").confirm() end, { desc = "Permission: confirm" })
vim.keymap.set("n", "<leader>op", function() require("opencode.events.permissions.floating").toggle() end, { desc = "Toggle permission window" })
```

## Testing

1. Ask agent to make multiple concurrent requests -> check if popup shows up unfocused and dim
2. Nav to popup -> should highlight + select 1st option
3. h/l should navigate options
4. n/p should navigate 
5. enter should accept option and changes should be reflected both in tui and nvim changes from tui should also be reflected in nvim
6. Esc/q while window is focused should hide it, until you toggle it back on from toggle() command
7. Allow all should accept all requests and not cause tool exception
8. repeat 1 but ask for multi line request which should be formatted correctly
9. repeat 1 but ask for edit, floating popup should not show up
10. repeat 1 and accept from tui, changes should be reflected in nvim floating window
11. All keybinds should work
12. 

## Related Issue(s)

Not reported as an issue but there was a discussion regarding a non intrusive solution to permissions in https://github.com/nickjvandyke/opencode.nvim/discussions/290#discussioncomment-17415758
I had already been making this locally and wanted to share in case you were interested in this

### DIFF
--- a/lua/opencode/config.lua
+++ b/lua/opencode/config.lua
@@ -107,6 +107,7 @@ local defaults = {
     reload = true,
     permissions = {
       enabled = true,
+      floating = false,
       edits = {
         enabled = true,
       },
--- a/lua/opencode/events/permissions/floating.lua
+++ b/lua/opencode/events/permissions/floating.lua
@@ -15,7 +15,7 @@ local M = {}
 -- State
 -- ──────────────────────────────────────────────
 
----@type {server: opencode.server.Server, queue: {event: table, id: string}[], queue_index: integer, win: integer|nil, buf: integer, width: integer, visible: boolean}|nil
+---@type {server: opencode.server.Server, queue: {event: table, id: number}[], queue_index: integer, win: integer|nil, buf: integer, width: integer, visible: boolean}|nil
 local state = nil
 
 local OPTIONS = { "Once", "Always", "Reject" }
@@ -77,7 +77,7 @@ local function pad_lines(lines, width)
 end
 
 local function calc_height(n_wrap)
-  return 1 + n_wrap + 1 + 1 + 1 + 1 + 1
+  return n_wrap + 6
 end
 
 ---Return the display prefix for a permission type.
@@ -414,7 +414,7 @@ function M.hide()
   state.win = nil
   state.visible = false
   vim.schedule(function()
-    if vim.api.nvim_win_is_valid(win) then
+    if win and vim.api.nvim_win_is_valid(win) then
       vim.api.nvim_win_close(win, true)
     end
   end)
--- a/lua/opencode/events/permissions/floating.lua
+++ b/lua/opencode/events/permissions/floating.lua
@@ -125,9 +125,9 @@ local function render(lines)
   local w = state.width
   pad_lines(lines, w)
 
-  vim.api.nvim_buf_set_option(state.buf, "modifiable", true)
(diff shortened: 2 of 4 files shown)


nickjvandyke:
Thank you for your contribution and effort put into this! However I don't think it fits the plugin's focus. Please see https://github.com/nickjvandyke/opencode.nvim/discussions/290#discussioncomment-17741329 for details. 🙂

PR #317, opened 2026-08-05
feat(rename): add session rename command

Alex-duzhichao:
- Track the TUI's active session via `tui.session.select` events
- Add `Server:update_session()` using `PATCH /session/:id`
- Route `session.rename` and `session.select` through `command()` instead of the TUI
- Expose `rename_session()` on the public API

## Tasks

## Testing

1. Run `:lua require("opencode").select()` to open select picker
2. Select the `session.rename`
3. Change the session name

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -6,7 +6,7 @@ A Neovim Lua plugin that bridges Neovim and the `opencode` CLI (external binary)
 
 ## Entrypoints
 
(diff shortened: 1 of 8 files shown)


nickjvandyke:
Thanks for your contribution! This seems reasonable to include. Nice solution to track the active session. That's been an issue historically and I didn't think of this! I'll take a closer look when I get the chance.

Does OpenCode emit `tui.session.select` when a new session is created? Wondering if we need to track more events. Also wondering how to handle when we first connect to the server. I presume we can't get the session at that time, and must hide/disable the command.

PR #327, opened 2026-09-12
fix(windows): use vim.json.decode to avoid E5560 in fast-event context

vad-bond:
## Tasks

## Description

On Windows, server discovery sometimes surfaced `E5560: Vimscript function "json_decode" must not be called in a fast event context`, which the surrounding `pcall` mislabeled as a JSON parse failure.

`vim.fn.json_decode` is a Vimscript function call, which Neovim forbids from the fast-event context used by the `vim.uv` timer/job callback during Windows process discovery. This PR replaces it with `vim.json.decode`, which is Lua-native, API-equivalent for this use, and callable from fast-event context — so the `E5560` error no longer occurs and parsing behavior is unchanged.

## Testing

1. On Windows, start an OpenCode server externally, then trigger opencode.nvim's server discovery.
2. Before this fix: an `E5560` notification appears instead of the process being discovered.
3. After this fix: discovery succeeds silently; `vim.json.decode` parses the PowerShell JSON output identically to `vim.fn.json_decode` (single-object and array forms).

## Related Issue(s)

N/A

## Screenshots/Videos

N/A


nickjvandyke:
Good find, thank you!

